PR #4, opened 2015-01-30
Change autogen.sh for better compatibility ...

juanmaneo:
Hello everyone,

Using autogen.sh script with the one given from your google's gtest framework offer better detection of tools like libtoolize and other GNU autotools ...
Is it possible to change it that way in the main branch ? Would be great.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -2,6 +2,11 @@
 rm -rf autom4te.cache
 aclocal -I m4
 autoheader
-libtoolize --copy
+if glibtoolize --version >/dev/null 2>/dev/null; then
+  LIBTOOLIZE=${LIBTOOLIZE:-glibtoolize}
+else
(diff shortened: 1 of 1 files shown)


dunn:
Or at the very least something like 
```bash
if glibtoolize --version >/dev/null; then
    LIBTOOLIZE=glibtoolize
fi
LIBTOOLIZE=${LIBTOOLIZE:-libtoolize}
```

invokr:
This is still an issue on OS X. Please modify autogen.sh to include a simple check for glibtoolize or merge this pull request.

adamnovak:
The inability to know that libtoolize should be glibtoolize (or even follow the environment's `$LIBTOOLIZE`) also messes up the build of ekg/vg on OS X, which includes snappy as a submodule.

PR #5, opened 2015-02-28
Update snappy-stubs-public.h.in

loverszhaokai:
Change to two blank


sesse:
Hi,

Unfortunately after the move to Github, we're not in a position where it's easy for us to take in patches (since what we'd need to merge them into our own repositories is not in place). We might make an exception if something particularly interesting comes along (it would be a significant chunk of manual work), but not for formatting changes like this. Thus, I'm afraid I'm going to have to close it; sorry.

loverszhaokai:
Get it!

PR #10, opened 2015-10-24
Use ptrdiff_t instead of ssize_t

jibsen:
The compression code seems pretty standard compliant, except for the two places ssize_t (POSIX) is used.

Using ptrdiff_t should allow the snappy plugin for squash to compile with MSVC. See quixdb/squash#145.

### DIFF
--- a/snappy.cc
+++ b/snappy.cc
@@ -95,7 +95,7 @@ size_t MaxCompressedLength(size_t source_len) {
 //    ababababababababababab
 // Note that this does not match the semantics of either memcpy()
 // or memmove().
-static inline void IncrementalCopy(const char* src, char* op, ssize_t len) {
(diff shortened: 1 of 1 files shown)


benvanik:
Poke poke owners - can this be merged?

sesse:
Hi,

In general, if your platform doesn't support a given type, that's the role of config.h to fix, not the code itself. MSVC has its own nonstandard type SSIZE_T which works well for this purpose; just add a typedef to your config.h.

jibsen:
Personally, I think using a type not in the C standard when there is a suitable type present, is needlessly limiting the portability of code. But at least there is a config.h to work around it.

Also, strictly speaking `ssize_t` is only guaranteed to support the range \[-1, SSIZE_MAX\] (it is not a signed `size_t`, but a `size_t` with one error value), so the use in `IncrementalCopyFastPath` could potentially be undefined.

PR #11, opened 2015-10-26
Fix compilation warning

kxepal:
```
snappy.cc: In member function ‘void snappy::SnappySinkAllocator::Flush(size_t)’:
snappy.cc:1336:38: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     for (int i = 0; i < blocks_.size(); ++i) {
```
This warning doesn't shows with current compilation options, but it does when you compile snappy with -Wall flag.

### DIFF
--- a/snappy.cc
+++ b/snappy.cc
@@ -1333,7 +1333,7 @@ class SnappySinkAllocator {
   void Flush(size_t size) {
     size_t size_written = 0;
     size_t block_size;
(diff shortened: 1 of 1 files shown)


cmumford:
@googlebot rescan

kxepal:
Thanks for ping, but this PR is no longer actual. And nowday there are far more warnings on master than it was in 2015 year, so there should be another solution.

PR #12, opened 2015-10-27
Add typedef for ssize_t for MSVC.

aloukissas:


Optiligence:
Funny, because there already is PR #10 which solves the problem in an IMO better way.

aloukissas:
either works :) but looks like the project isn't monitored really by its committers

sesse:
Actually it is, but it's hard to accept pull requests the way things work internally right now (it's more or less a one-way process).

In any case, the typedef from SSIZE_T to ssize_t should be part of config.h, not snappy.cc. Maybe at some point we'll ship a config.h for MSVC, but it's tricky to maintain something on a platform you don't use. See the MSVC ports linked from the home page (e.g. https://snappy.angeloflogic.com/).

aloukissas:
suggest switching to CMake

Optiligence:
I figured. 
But you can do it like you do it right now, respond and close internally implemented or rejected PRs. 
PR #10 is still better than platform dependent branching though. 
+1 for CMake

aloukissas:
deal - i can close this PR and hopefully #10 gets merged soon

pwnall:
ssize_t is no longer used, so we removed the definitions around it. Thank you for your contribution, and sorry we didn't get to it in time!

PR #17, opened 2016-02-26
Using glibtoolize for OS X

BlueCocoa:
See [installed libtool but libtoolize not found](http://stackoverflow.com/questions/15448582/installed-libtool-but-libtoolize-not-found/15448876#15448876) on stackoverflow.


sesse:
Hi,

I just pushed a change similar to this, so this PR should be obsolete.

I'm sorry it took forever—there are excuses, but none of them good.

PR #18, opened 2016-03-22
Add support for unaligned loads/stores for ARM.

soumitrachatterjee:


sesse:
Hi,

In your description, I assume you mean ARM64? There's already support for unaligned loads/stored for 32-bit ARM in the code.

As for __ARM_FEATURE_UNALIGNED, what sets this? It wasn't set by any compilers last time I tested, so I don't think you can rely on it.

soumitrachatterjee:
@sesse, the changes are for both ARM and ARM64.

For ARM64, this enables the support for unaligned load/store.
For ARM, this is an attempt to clean up the code by avoiding all ARM CPU models in the code.

As regards `__ARM_FEATURE_UNALIGNED`, please take a look at article 6.4.3 in the [ACLE](http://infocenter.arm.com/help/topic/com.arm.doc.ihi0053b/IHI0053B_arm_c_language_extensions_2013.pdf).

As regards support for the same in gcc, please refer:
https://gcc.gnu.org/ml/gcc-patches/2011-09/msg00878.html
http://lists.gnupg.org/pipermail/gcrypt-devel/2013-August/002308.html

This is also used in LZO:
https://lkml.org/lkml/2013/2/3/367

sesse:
Yes, unfortunately GCC support from late 2011 (well, that's only the commit; probably released in GCC 4.7 in March 2012) is a tad too late for some platforms; embedded toolchains tend to update slowly. :-/

The effects are not catastrophic, though, the code just goes slower. Maybe it's time to draw the line, although e.g. Ubuntu 12.04 (still popular on various ARM platforms) still ships with only 4.6 by default.

soumitrachatterjee:
Given the feedback, I guess its better if I split the patch into two separate ones, one addressing the ARM64 support and the other for the cleanup. That should help at least the ARM64 support to be added, if not the other.

Closing this. I'll raise separate pull requests for both.

PR #20, opened 2016-05-16
Use "= delete" to implement DISALLOW_COPY_AND_ASSIGN.

pkasting:
This also adds an #ifdef to guard against macro redefinition if this is included in another Google project that also defines this.

### DIFF
--- a/snappy-stubs-public.h.in
+++ b/snappy-stubs-public.h.in
@@ -80,9 +80,11 @@ typedef unsigned long long uint64;
 
 typedef std::string string;
 
+#ifndef DISALLOW_COPY_AND_ASSIGN
 #define DISALLOW_COPY_AND_ASSIGN(TypeName) \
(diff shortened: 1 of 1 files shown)


sesse:
Hi,

This would introduce a dependency on C++11 for what's essentially a no-op, so sorry, no.

pkasting:
You don't already depend on C++11?

Would you be willing to take the #ifdef part, then?  This redefinition is causing problems in Chromium, which does want to use "= delete" for its definition of DISALLOW_COPY_AND_ASSIGN.

pkasting:
Opened https://github.com/google/snappy/pull/21 for just the #ifdef.

sesse:
No, there's no C++11 dependency, it's all C++98.

Having an #ifdef would be fine; if a macro with such a name is #defined, it's very likely to do the right thing.

PR #21, opened 2016-05-16
Add #ifdef around definition of DISALLOW_COPY_AND_ASSIGN.

pkasting:
This prevents macro redefinition errors when other Google projects (e.g. Chromium) attempt to use Snappy.


sesse:
Merged (after squash), thanks.

PR #24, opened 2016-10-26
Fix various compiler warnings

eile:
Found while integrating in https://github.com/eile/Pression/commit/26c4d546037ba1236bbdfd9fe2f84d3092bf7e4d

We're using CMake as a build system, and I'm compiling snappy directly
from there since autoconf integration would be painful. We've got fairly
aggressive warnings on, and this PR fixes all which appeared in our
setting. IIRC a CLA for me should already be in the system from previous
contributions to another google project.

### DIFF
--- a/snappy.cc
+++ b/snappy.cc
@@ -1358,11 +1358,12 @@ public:
   inline SnappyDecompressionValidator() : expected_(0), produced_(0) {}
   inline void SetExpectedLength(size_t len) { expected_ = len; }
   inline bool CheckLength() const { return expected_ == produced_; }
-  inline bool Append(const char *ip, size_t len) {
+  inline bool Append(const char * /*ip*/, size_t len) {
     produced_ += len;
     return produced_ <= expected_;
   }
-  inline bool TryFastAppend(const char *ip, size_t available, size_t length) {
+  inline bool TryFastAppend(const char * /*ip*/, size_t /*available*/,
+                            size_t /*length*/) {
     return false;
   }
   inline bool AppendFromSelf(size_t offset, size_t len) {
@@ -1588,7 +1589,7 @@ private:
     Datablock(char *p, size_t s) : data(p), size(s) {}
   };
 
-  static void Deleter(void *arg, const char *bytes, size_t size) {
+  static void Deleter(void * /*arg*/, const char *bytes, size_t /*size*/) {
     delete[] bytes;
   }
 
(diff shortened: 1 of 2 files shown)


VVD:
Sorry for offtopic, but I can't find any issue/bug tracker, mailing list or developer's emails of this project (snappy) for report a bug and patch.
Snappy 1.1.4 fails to compile on i386 with -march=prescott, or core2 and etc.
Patch is very simply:
```
--- snappy.cc.orig        2017-01-30 12:03:09
+++ snappy.cc
@@ -30,7 +30,7 @@
 #include "snappy-internal.h"
 #include "snappy-sinksource.h"
 
-#if defined(__x86_64__) || defined(_M_X64)
+#ifdef __SSE2__
 #include <emmintrin.h>
 #endif
 #include <stdio.h>
```
More details are at FreeBSD's bugzilla: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=216553

Thanks!

alkis:
Thanks for the bug report. I will work on a fix.

grooverdan:
i386 fixed 626e1b9faa81e6f744b8b101ab0cf94c5927d15a. For my own benefit can unused parameters be correct by attribute too?

cmumford:
@googlebot rescan

snappy.cc
```diff
@@ -1245,7 +1245,7 @@ class SnappyScatteredWriter {
   inline bool AppendFromSelf(size_t offset, size_t len) {
     // See SnappyArrayWriter::AppendFromSelf for an explanation of
     // the "offset - 1u" trick.
-    if (offset - 1u < op_ptr_ - op_base_) {
+    if (offset - 1u < size_t(op_ptr_ - op_base_)) {
```
cmumford: Can you switch casts to [C++-style casts ](https://google.github.io/styleguide/cppguide.html#Casting)?

snappy.cc
```diff
@@ -670,7 +670,7 @@ bool SnappyDecompressor::RefillTag() {
       size_t length;
       const char* src = reader_->Peek(&length);
       if (length == 0) return false;
-      uint32 to_add = min<uint32>(needed - nbuf, length);
+      uint32 to_add = std::min<uint32_t>(needed - nbuf, length);
```
cmumford: Can you switch the type from uint32 to uint32_t as well?

cmumford (changes requested):
Thx for the PR @eile. Minor changes requested. Can you also run clang-format?

eile:
@cmumford done.

PR #26, opened 2017-02-06
Revert "Improve zippy decompression speed."

ncopa:
This reverts commit 8bfb028b618747a6ac8af159c87e9196c729566f.

The commit 8bfb028 (Improve zippy decompression speed) introduce a crash
when snappy is compiled with _FORTIFY_SOURCE with musl libc. Backtrace
reveals that it it comes from using memcpy with overlap. Since this may
have security implications we better revert it for now.

Bactrace from core dump created with `make check`:
````
(gdb) bt
 #0  memcpy (__n=8, __os=0xb38c8367eea, __od=0xb38c8367eeb)
     at /usr/include/fortify/string.h:48
 #1  snappy::(anonymous namespace)::UnalignedCopy64 (src=0xb38c8367eea,
     dst=0xb38c8367eeb) at snappy.cc:92
 #2  0x00006fb4c7c31717 in snappy::(anonymous namespace)::IncrementalCopy
 (
     buf_limit=0xb38c8380ee0 "", op_limit=<optimized out>, op=<optimized
 out>,
     src=0xb38c8367eea " .\001") at snappy.cc:178
 #3  snappy::SnappyArrayWriter::AppendFromSelf (len=<optimized out>,
     offset=<optimized out>, this=<synthetic pointer>) at snappy.cc:1131
 #4
 snappy::SnappyDecompressor::DecompressAllTags<snappy::SnappyArrayWriter>
 (
     writer=<synthetic pointer>, this=0x7f1d26737050) at snappy.cc:715
 #5  snappy::InternalUncompressAllTags<snappy::SnappyArrayWriter> (
     uncompressed_len=<optimized out>, writer=<synthetic pointer>,
     decompressor=0x7f1d26737050) at snappy.cc:799
 #6  snappy::InternalUncompress<snappy::SnappyArrayWriter> (
     writer=<synthetic pointer>, r=0x7f1d26737000) at snappy.cc:789
 #7  snappy::RawUncompress (compressed=compressed@entry=0x7f1d267370c0,
     uncompressed=0xb38c8367ee0 "  content: .\001") at snappy.cc:1149
 #8  0x00006fb4c7c3194d in snappy::RawUncompress (compressed=<optimized
 out>,
     n=<optimized out>, uncompressed=<optimized out>) at snappy.cc:1144
 #9  0x00000b38c5c6261a in snappy::BM_UFlat (iters=99, arg=<optimized
 out>)
     at snappy_unittest.cc:1371
 #10 0x00000b38c5c69846 in snappy::Benchmark::Run (this=0xb38c8323c40)
     at snappy-test.cc:192
 #11 0x00000b38c5c5f3fd in RunSpecifiedBenchmarks () at snappy-test.h:485
 #12 main (argc=1, argv=0x7f1d267374b8) at snappy_unittest.cc:1515
````

### DIFF
--- a/snappy-stubs-internal.h
+++ b/snappy-stubs-internal.h
@@ -225,20 +225,6 @@ inline void UNALIGNED_STORE64(void *p, uint64 v) {
 
 #endif
 
-// This can be more efficient than UNALIGNED_LOAD64 + UNALIGNED_STORE64
-// on some platforms, in particular ARM.
-inline void UnalignedCopy64(const void *src, void *dst) {
-  if (sizeof(void *) == 8) {
-    UNALIGNED_STORE64(dst, UNALIGNED_LOAD64(src));
-  } else {
-    const char *src_char = reinterpret_cast<const char *>(src);
-    char *dst_char = reinterpret_cast<char *>(dst);
-
-    UNALIGNED_STORE32(dst_char, UNALIGNED_LOAD32(src_char));
-    UNALIGNED_STORE32(dst_char + 4, UNALIGNED_LOAD32(src_char + 4));
-  }
-}
-
 // The following guarantees declaration of the byte swap functions.
 #ifdef WORDS_BIGENDIAN
 
--- a/snappy.cc
+++ b/snappy.cc
@@ -89,7 +89,9 @@ size_t MaxCompressedLength(size_t source_len) {
 namespace {
 
 void UnalignedCopy64(const void* src, void* dst) {
-  memcpy(dst, src, 8);
+  uint64 tmp;
+  memcpy(&tmp, src, 8);
+  memcpy(dst, &tmp, 8);
 }
 
 void UnalignedCopy128(const void* src, void* dst) {

### CODE REVIEW

**Comment by ncopa** [comment]
I wasn't able to find a bugtracker so I made a PR.

I have not analyzed if it is exploitable, but it is very possible that it is and that a CVE should be requested.

Another possible fix is to replace `memcpy` with `memmove`:
````diff
diff --git a/snappy.cc b/snappy.cc
index 4bcea0b..2db94b7 100644
--- a/snappy.cc
+++ b/snappy.cc
@@ -89,7 +89,7 @@ size_t MaxCompressedLength(size_t source_len) {
 namespace {
 
 void UnalignedCopy64(const void* src, void* dst) {
-  memcpy(dst, src, 8);
+  memmove(dst, src, 8);
 }
 
 void UnalignedCopy128(const void* src, void* dst) {
````

**Comment by alkis** [comment]
There is no CVE here. Wherever UnaligedCopy64 is used, it assumes that memory might be overlapping and the overlapping bytes are OK to be clobbered. This assumption is verified by the calling code. Perhaps you can suggest a way to circumvent this false positive?

**Comment by ncopa** [comment]
The `memcpy` [specification](http://pubs.opengroup.org/onlinepubs/9699919799/functions/memcpy.html) explicitly says:
> If copying takes place between objects that overlap, the behavior is undefined.

The [compiler can rightfully just drop the entire code path](https://en.wikipedia.org/wiki/Undefined_behavior#Benefits_from_undefined_behavior) when the arguments overlap, or compiler can rightfully decide to just crash, which happened in my hardened toolchain. There is also no guarantee that it is the overlapping region that is clobbered and not the non-overlapping region, and you can not guarantee that anything is copied at all.

There is basically no guarantee for anything when the behavior is undefined, and it is always [risky](https://en.wikipedia.org/wiki/Undefined_behavior#Risks_of_undefined_behavior) to rely on it.

I did work around it on Alpine Linux by replacing the `memcpy` with `memmove`.

Feel free to reject and close.

**Comment by amonakov** [comment]
You can use a suitably sized temporary to get well-optimized code without undefined behavior:
```c
void UnalignedCopy64(const void* src, void* dst) {
  int64_t tmp;
  memcpy(&tmp, src, sizeof tmp);
  memcpy(dst, &tmp, sizeof tmp);
}
```

sesse:
Or you can just do what the code did before the patch, which has the added advantage that you don't get the pass ordering problem that memcpy has (the memcpy inlining happens late enough that it sometimes inhibits other optimizations):

```c
    UNALIGNED_STORE64(dst, UNALIGNED_LOAD64(src));
```

amonakov:
A problem with existing unaligned-access macros in snappy is that they hide from the compiler that the accesses may be unaligned (which also leads to undefined behavior). GCC and Clang support `attribute((aligned(N)))` to declare objects with reduced alignment, which can be used here:
```c
void UnalignedCopy64(const void* src, void* dst) {
  typedef int64_t ui64 __attribute__((aligned(1)));
  int64_t tmp = *(const ui64*)src;
  *(ui64*)dst = tmp;
}
```

sesse:
> A problem with existing unaligned-access macros in snappy is that they hide from the compiler that the accesses may be unaligned (which also leads to undefined behavior). GCC and Clang support attribute((aligned(N))) to declare objects with reduced alignment, which can be used here:

The UNALIGNED_* macros already use this on relevant platforms. You could argue that it should be extended to x86 too, of course, but that's a different discussion.

Apart from that, I'm not sure why the patch introduced a second UnalignedCopy64() function; there's already one in snappy-stubs-internal.h that has worked well for a long time.

alkis:
I think the best solution here is to do:

```c++
void UnalignedCopy64(const void* src, void* dst) {
  uint64 tmp;
  memcpy(&tmp, src, 8);
  memcpy(dst, &tmp, 8);
}
```

The UnalignedCopy64() in snappy-stubs-internal.h is strictly worse than the above.

As for the UNALIGNED_* macros, I don't understand why they even exist. UnalignedLoad64() should just memcpy into uint64 and return. The compiler knows how to most efficiently load 8 bytes of memory into a 64 bit register.

sesse:
> The UnalignedCopy64() in snappy-stubs-internal.h is strictly worse than the above.

Except that it is faster.

> As for the UNALIGNED_* macros, I don't understand why they even exist. UnalignedLoad64() should just memcpy into uint64 and return. The compiler knows how to most efficiently load 8 bytes of memory into a 64 bit register. 

In an ideal world, it would, but alas. Again, the lowering of memcpy() to a simple load happens pretty late in  the compilation, so you're missing out on other optimizations in the code around it. It's a classic pass ordering issue; you can't just move it arbitrarily early either, since you need it after e.g. constant folding if someone does memcpy(foo, bar, 4+4). (Since I don't work in Google anymore, I obviously don't have access to the mailing list archives, but it was discussed on c-compiler-opt@ when we benchmarked changing UNALIGNED_* to memcpy and noticed it gave worse speed.)

alkis:
Was the thread about snappy specifically? The one I found was about a summing loop.

In the particular case of snappy I changed UnorderedCopy64() to:

```c++
void UnalignedCopy64(const void* src, void* dst) {
  uint64 tmp;
  memcpy(&tmp, src, 8);
  memcpy(dst, &tmp, 8);
}
```
Benchmarking against:
```c++
void UnalignedCopy64(const void* src, void* dst) {
  memcpy(dst, src, 8);
}
```
Shows no difference.

sesse:
> Was the thread about snappy specifically? The one I found was about a summing loop.

I can't recall, sorry; this is a few years ago now and I'm not sure if I mentioned Snappy per se.

> In the particular case of snappy I changed UnorderedCopy64() to:

I'm not talking about the difference between one memcpy and two memcpys, but the difference between memcpy and the UNALIGNED_* macros (which you said you “don't understand why even exist”).

IIRC someone suggested we change UNALIGNED_* to generally just use memcpy on the same basis, and it yielded worse code (something on the order of 5% on Snappy). Of course, it works great in a microbenchmark on its own, because these effects only happen when the compiler also need to do other, complex optimizations to the surrounding code. And again, there's certainly cases where you get lucky and don't hit these issues. I doubt we tried it on a larger Google3 suite, given that it we had so clear indications that life wasn't so simple as “use memcpy, it will be just as fast as regular accesses”.

alkis:
I tested
```c++
void UnalignedCopy64(const void* src, void* dst) {
  char tmp[8];
  memcpy(&tmp, src, 8);
  memcpy(dst, &tmp, 8);
}
```

vs

```c++
void UnalignedCopy64(const void* src, void* dst) {
  UNALIGNED_STORE(dst, UNALIGNED_LOAD(src));
}
```

And there was no measurable difference (old is `memcpy`, new is `UNALIGNED_*`):

```
name         old speed      new speed      delta
BM_UFlat/0   3.14GB/s ± 1%  3.16GB/s ± 2%    ~     (p=0.156 n=19+19)
BM_UFlat/1   1.62GB/s ± 1%  1.61GB/s ± 2%    ~     (p=0.102 n=19+20)
BM_UFlat/2   18.8GB/s ±17%  19.1GB/s ±11%    ~     (p=0.390 n=20+16)
BM_UFlat/3   1.59GB/s ± 1%  1.58GB/s ± 1%  -1.06%  (p=0.000 n=18+18)
BM_UFlat/4   15.8GB/s ± 6%  15.6GB/s ± 7%    ~     (p=0.184 n=19+20)
BM_UFlat/5   2.46GB/s ± 1%  2.44GB/s ± 1%  -0.95%  (p=0.000 n=19+18)
BM_UFlat/6   1.08GB/s ± 1%  1.06GB/s ± 1%  -1.17%  (p=0.000 n=19+18)
BM_UFlat/7   1.00GB/s ± 1%  0.99GB/s ± 1%  -1.16%  (p=0.000 n=19+18)
BM_UFlat/8   1.14GB/s ± 2%  1.12GB/s ± 1%  -1.12%  (p=0.000 n=19+18)
BM_UFlat/9    921MB/s ± 1%   914MB/s ± 1%  -0.84%  (p=0.000 n=20+17)
BM_UFlat/10  3.94GB/s ± 2%  3.92GB/s ± 1%    ~     (p=0.058 n=19+17)
BM_UFlat/11  1.29GB/s ± 1%  1.28GB/s ± 1%  -0.77%  (p=0.001 n=19+17)
BM_UFlat/12  2.34GB/s ± 1%  2.31GB/s ± 1%  -1.10%  (p=0.000 n=18+18)
BM_UFlat/13  2.06GB/s ± 1%  2.05GB/s ± 1%  -0.73%  (p=0.001 n=19+18)
BM_UFlat/14  2.22GB/s ± 1%  2.20GB/s ± 1%  -0.73%  (p=0.000 n=18+18)
BM_UFlat/15  1.55GB/s ± 1%  1.53GB/s ± 1%  -1.07%  (p=0.000 n=19+18)
BM_UFlat/16  1.26GB/s ± 1%  1.25GB/s ± 1%  -0.79%  (p=0.000 n=18+18)
BM_UFlat/17  2.31GB/s ± 1%  2.29GB/s ± 1%  -0.98%  (p=0.000 n=20+18)
BM_UFlat/18  1.61GB/s ± 1%  1.60GB/s ± 2%  -0.71%  (p=0.001 n=20+19)
BM_UFlat/19  1.77GB/s ± 1%  1.76GB/s ± 1%  -0.61%  (p=0.007 n=19+18)
BM_UFlat/20  1.89GB/s ± 1%  1.88GB/s ± 1%  -0.75%  (p=0.000 n=20+18)
```

alkis:
Fixed by 3c706d22306efbd0d6ac050cb48179e1dcfbff7a.

PR #30, opened 2017-04-03
Support MSVC _BitScan intrinsics

KindDragon:


KindDragon:
@alkis Any news about this PR?

alkis:
This didn't make the cut. It will be in the next release.

pwnall:
This was landed in f0d3237c320fb4b3e5d9d05c7182f33fdca7abe6. Thank you very much!

KindDragon:
@pwnall Why not implement `FindLSBSetNonZero64` for 32-bit platform?

pwnall:
@KindDragon `FindLSBSetNonZero64` is not called in 32-bit builds. The proof is that AppVeyor is set up to do 32-bit Visual Studio builds, and the CI is green.

Under these circumstances, implementing the method has two downsides: We'd end up maintaining code that's not used, and we'd ship dead code on systems where the linker doesn't do dead code removal.

PR #31, opened 2017-04-06
Update index.html to include MongoDB

Dabz:
MongoDB is using Snappy compression by default in its wiredTiger storage engine. I think it's worth mentioning (https://docs.mongodb.com/manual/core/wiredtiger/#compression)


Dabz:
I signed it!

PR #33, opened 2017-04-12
to resolve "/usr/bin/install: cannot stat './README': No such file or…

imriss:
To resolve the error when installing: "/usr/bin/install: cannot stat './README': No such file or directory"


imriss:
I signed it!

chakree1g:
I signed it!

pwnall:
This has the same fate as https://github.com/google/snappy/pull/46. We removed autoconf support, so we don't need to worry about the README workaround anymore. Thank you for your contribution, and sorry we didn't get to it earlier!

PR #36, opened 2017-04-23
fix a signed/unsigned comparison warning

eklitzke:
This fixes a minor error, where a signed integer is used to in a loop instead of a `size_t`. This fixes a compiler warning when building snappy.


eklitzke:
ping

cmumford:
@googlebot rescan

PR #37, opened 2017-05-13
Make SSE2 check compatible with MSVC

KindDragon:


alkis:
Closing this as it will be part of the next release.

PR #29, opened 2017-02-27
Allow building with cmake (to ease building on windows), version 2

vrabaud:
This is a small rewrite of #16 , mostly to simplify the windows support and modify less code.

PR #38, opened 2017-06-06
CMake improvements

Optiligence:
So i tried to get my CMake wishes into the previous PRs, some of it landed, this is the rest. 

Since the changed CMake config is pretty important for downstream use, i hope this can be included before the initial CMake release. 

https://github.com/google/snappy/pull/38/commits/ad817785de16d59c6bc6a0a3ca379ca145453d2c is my take on one of CMake’s weak points, autotools allowed to do this before.


Optiligence:
I signed it!

CMakeLists.txt
```diff
@@ -81,7 +81,7 @@ if (WIN32)
 endif (WIN32)
 
 # Define the main library.
-add_library(snappy SHARED
+add_library(Snappy SHARED
```
mgorny: Why? This is something other packages may have already started relying on, and I really see no advantage of making it uppercase (and therefore inconsistent with `-lsnappy`).

Even worse, if you change this, then packages that already `target_link_libraries(snappy)` will silently switch from linking the target to implicit `-lsnappy`.

Optiligence: https://github.com/Optiligence/snappy/commit/fb66b509cdb31687414654f158ac4617d18bc1af#diff-af3b638bc2a3e6c650974192a53c7291L96
The output name stays the same. 
CMake targest are usuaply upper cased.

ilovezfs:
@pwnall we just had a user report the lack of a static library: https://github.com/Homebrew/homebrew-core/issues/15722

Optiligence:
rebased again

CMakeLists.txt
```diff
@@ -101,12 +103,17 @@ install(FILES snappy.h
         ${Snappy_BINARY_DIR}/snappy-stubs-public.h
         DESTINATION include)
 
+if (BUILD_SHARED_LIBS)
```
dougalsutherland: Wouldn't it be nicer to have both `BUILD_SHARED_LIBS` and `BUILD_STATIC_LIBS`, so you can build both a static and a shared library without reconfiguring in between?

Optiligence: There is no `BUILD_STATIC_LIBS`, `BUILD_SHARED_LIBS` [is a CMake variable](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html). 
You would have to add 2 libraries and create targets with different names for each
combined with custom logic with enables or disables it. 
downstream projects can’t use both at the same time anyway
The way it is now downstream gets shared snappy by default and would get static snappy when BUILD_SHARED_LIBS is set (which i consider the build-all-my-stuff-statically-please flag) which has the benefit that downstream doesn’t have to explicitly state anything in their CMake file, it’s only a build option. 
One could also add explicit aliases on top of that (in the future?). 
Reconfiguring in between has no downside anyway, it’s just a stylistic issue. 
I’m quite happy with the way it is now, I also don’t know of any CMake projects which handle building static and shared flavors nicely to compare it with.

pwnall:
Generic advice: If you want your PR to be helpful, please don't bundle a coding style change with an actual change.

In general, style choices tend to be intentional, and are (rightfully) optimized for the project maintainers, who'll have to stare at the code for extended periods of time. In this case, I agree with the style change, as it matches the style in the CMake documentation, but the bundling still makes the diff useless.

I plan to look into this issue today.

Optiligence:
When I initially brought this up [here](https://github.com/google/snappy/pull/29#issuecomment-288733188) I maintained 2 versions, one with the capitalization changed and one without. 
Since this got annoying pretty quick (after a few rebases without anybody responding) I concentrated on the version I was more comfortable with. 
Imo it’s also not really useful to look at the complete diff anyway, the individual commits are easier to reason about (especially together with the commit message). 

Thanks for looking into it.

CMakeLists.txt
jakirkham: I think most CMake projects end up adding a `_static` suffix to the library name. Honestly this is actually pretty nice as one can clearly specify they want static linking on Unix by simply using this other library name.

pwnall:
@Optiligence The project's CMake configuration has been revamped. Sorry to ask for more work from you, but can you please check which parts of this PR still apply and rebase?

Also, if individual commits are easier to reason about, that's a pretty good sign that the commits should be in separate PRs.

mgorny:
Thank you for rewriting a crappy build system into another equally crappy build system without fixing a single issue and requiring us to write all the patches over again just to make it remotely useful.

ilovezfs:
@mgorny Please try to be nice.

Optiligence:
I will work on it when i find the time.

Optiligence:
@pwnall You want ALLCAPS now or no?

pwnall:
@Optiligence no more CAPS :smile:  I hope that'll make diff management easier for you.

pwnall:
@Optiligence Also, please make sure to rebase against master, not the latest release.

Optiligence:
Oh, nice, I didn’t have a look at the latest commit yet.

Optiligence:
Those 3 commits seem to be implemented: https://github.com/google/snappy/pull/38/commits/66bd85143187052b7e59c8dc1f8b037f8c4100b7 https://github.com/google/snappy/pull/38/commits/e23726b936392685df5de6791dc6af317e5551f1 https://github.com/google/snappy/pull/38/commits/26adadf5301f0f8bb2a60fc48798a11895dd8a5a
I will create a new PR (https://github.com/google/snappy/pull/54) for the rest because it’s not possible to rebase reasonably anymore. 

On another note: 
Reading and managing multiple interdependent PRs is not easier but more difficult than reading individual commits in 1 PR. (in other words: if individual commits are dependent on each other, that’s a pretty good sign that they should belong to 1 PR)
https://github.com/google/snappy/commit/be6dc3db83c4701e3e79694dcbfd1c3da03b91dd is also pretty interesting regarding the comment about unreadable merge commits (which is part of the nature of merge commits but not standalone commits). ^^

PR #39, opened 2017-06-25
Modify snappy test Windows.h include to not define windows specific min/max

BrennenTaylor:
This was causing a compilation bug with VS 2017 Win64 as the Windows.h MAX macro conflicted with the use of std::max.

### DIFF
--- a/snappy-test.cc
+++ b/snappy-test.cc
@@ -33,6 +33,7 @@
 #endif
 
 #ifdef HAVE_WINDOWS_H
+#define NOMINMAX
 #include <windows.h>
 #endif
 


BrennenTaylor:
I signed it!

pwnall:
I fixed this in https://github.com/google/snappy/commit/e9720a001d4feffe1fc9770859fe9913d0d5ede6 which has been in the pipeline for a bit of time. Sorry for the delay in getting it out!

BrennenTaylor:
Gotcha!! No worries, thanks for getting that change in! :)

PR #42, opened 2017-06-29
Error for missing README

gdsotirov:
Should not this rule produce README, instead of README.tmp? Otherwise, I receive the following error:

```
cat README.md > README.tmp
 /usr/bin/mkdir -p '/usr/src/tmp/package-snappy/usr/doc/snappy-1.1.5'
 /usr/bin/ginstall -c -m 644 ChangeLog COPYING INSTALL NEWS ./README format_description.txt framing_format.txt '/usr/src/tmp/package-snappy/usr/doc/snappy-1.1.5'
/usr/bin/ginstall: cannot stat './README': No such file or directory
Makefile:767: recipe for target 'install-dist_docDATA' failed
make[1]: *** [install-dist_docDATA] Error 1
make[1]: Leaving directory '/usr/src/tmp/snappy-1.1.5'
Makefile:1230: recipe for target 'install-am' failed
make: *** [install-am] Error 2
```

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -28,4 +28,4 @@ libtool: $(LIBTOOL_DEPS)
 # Needed by autoconf because we use README.md instead of README.
 # See http://stackoverflow.com/q/15013672/
 README: README.md
-	cat $< > $@.tmp
+	cat $< > $@


gdsotirov:
I signed it!

gdsotirov:
The problem with Autotools in 1.1.5 is also that SO version is 0.4.1 (i.e. `libsnappy.so.0.4.1`), which breaks existing run-time dependencies that relay on `libsnappy.so.1` link to be present.

Makefile.am
```diff
@@ -28,4 +28,4 @@ libtool: $(LIBTOOL_DEPS)
 # Needed by autoconf because we use README.md instead of README.
 # See http://stackoverflow.com/q/15013672/
 README: README.md
```
bjori: shouldn't this rule be removed and the README in dist_doc_DATA be renamed README.md?

gdsotirov: Autoconf and Linux distributions require just README (without .md, .txt or other extension).

dougalsutherland:
:+1: Needed to add this to https://github.com/conda-forge/snappy-feedstock/pull/11 to build snappy there.

Makefile.am
bjori: My bad. autogen.sh would also need to be updated to pass `--foreign` to automake to stop it from complaining about the names.
These are very silly warnings anyway.

pwnall:
Thank you very much for the PR! We "fixed the glitch" by removing autoconf support. I hope CMake will be a less bumpy ride, once we settle on a configuration that works for everyone.

gdsotirov:
Sure, but please, consider adding the option to set LIBDIR with CMake, because it's important.

pwnall:
@gdsotirov Can you please create a PR or comment on the relevant PR?

gdsotirov:
@pwnall I do not have a patch for LIBDIR support, because I'm not that familiar with CMake yet, so I've solved the problem by just renaming the directory after `make install` :-) Still it's good to consider for next version. Where can I raise an issue, because here on GitHub I do not see this option?

pwnall:
@gdsotirov I think we don't have Issues enabled because we're struggling to just keep up the pace with the PRs... at this rate, issues would just get ignored. Let's continue the discussion here for now -- I thought LIBDIR was discussed somewhere in the open PRs, but that doesn't seem to be the case.

I'm still catching up on CMake, and I'm wary of adding unnecessary complexity to our CMakeLists. To this end, I'd be more interested in the reasoning behind a change, and the sources advocating for it.

In this specific example, I'd like to know what problems are solved by exposing a LIBDIR variable, and why that's not advocated in [CMake's packaging guide](https://cmake.org/cmake/help/git-master/manual/cmake-packages.7.html). Yesterday, the [GNUInstallDirs](https://cmake.org/cmake/help/git-master/module/GNUInstallDirs.html) was brought to my attention. It seems like it'd solve the LIBDIR problem, but I haven't had time to look into why it's not recommended in the packaging guide.

gdsotirov:
@pwnall OK about issues.

Exposing the possibility to set LIBDIR is important for distribution packages like me, because the library directory for a true 64 bit OS like Slackware64 is `/usr/lib64`, so I need to set it properly for installation. With autotools it was possible to set LIBDIR (e.g. I was managing it with `./configure --libdir=/usr/lib${LIBDIRSUFFIX}`), so I believe the lack of possibility with CMake is a regression, that needs to be fixed. I build other library packages with CMake and they all provide the possibility to set LIBDIR.

Optiligence:
What do other CMake projects use to achieve this, do they use `GNUInstallDirs`, specify custom variables themselves, or something else? 
I don’t like that `DESTINATION` has to be manually set for `install(TARGETS)` and doesn’t have sane (and configurable) defaults instead.

gdsotirov:
BTW I see Gentoo packages have already elaborated on such patch (see [snappy-1.1.6-0001-cmake-Use-GNUInstallDirs-to-provide-configurable-cor.patch](https://gitweb.gentoo.org/repo/gentoo.git/tree/app-arch/snappy/files/snappy-1.1.6-0001-cmake-Use-GNUInstallDirs-to-provide-configurable-cor.patch)). And also the same patch as my PR #44 (see https://gitweb.gentoo.org/repo/gentoo.git/tree/app-arch/snappy/files/snappy-1.1.6-0003-cmake-Add-missing-linking-to-GTEST_LIBRARIES.patch). I still believe that both are necessary, but as I'm not that familiar with CMake I haven't suggested a fix for LIBDIR.

gdsotirov:
@Optiligence I cannot say, but for example for [libproxy](https://github.com/libproxy/libproxy) I use `cmake -DLIB_INSTALL_DIR=...` and for [libgme](https://bitbucket.org/mpyne/game-music-emu/) its `cmake -DLIB_SUFFIX=...`, so you could check what they use if you want.

pwnall:
This is also being discussed in https://github.com/google/googletest/pull/1141. I'm heavily inclined to adopt googletest's decision for fixing this problem.

Optiligence:
The missing `RUNTIME DESTINATION`, `LIBRARY DESTINATION` und `ARCHIVE DESTINATION` also currently install the `*.dll` to the lib directory. +1 for GNUInstallDirs.

gdsotirov:
Whatever gives me the possibility to set properly LIBDIR is OK for me and Gentoo packages have already implemented GNUInstallDirs, so +1 for GNUInstallDirs from me as well :-)

pwnall:
@gdsotirov @Optiligence Can you please check `master` and see if the GNUInstallDirs solution meets your needs?

gdsotirov:
@pwnall Sorry, for the delay. I just checked the freshly released [Snappy 1.1.7](https://github.com/google/snappy/releases/tag/1.1.7) and I confirm it's OK with the new option CMAKE_INSTALL_LIBDIR. Thanks!

pwnall:
@gdsotirov Thank you very much for the response!

PR #44, opened 2017-07-03
Add GTEST_BOTH_LIBRARIES for link target snappy-unittest

gdsotirov:
It's necessary to add ${GTEST_BOTH_LIBRARIES} for link target snappy-unittest to avoid errors about missing GTest symbols. Also I'm not sure why FUND_PACKAGE for GTest and Gflags is with option QUIET, because it's better to know if anything is missing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,12 +56,12 @@ CHECK_CXX_SOURCE_COMPILES("int main(void) { return __builtin_expect(0, 1); }"
 CHECK_CXX_SOURCE_COMPILES("int main(void) { return __builtin_ctzll(0); }"
         HAVE_BUILTIN_CTZ)
 
-FIND_PACKAGE(GTest QUIET)
+FIND_PACKAGE(GTest)
 IF(GTEST_FOUND)
     SET(HAVE_GTEST 1)
 ENDIF()
 
-FIND_PACKAGE(Gflags QUIET)
+FIND_PACKAGE(Gflags)
 IF(GFLAGS_FOUND)
     SET(HAVE_GFLAGS 1)
 ENDIF()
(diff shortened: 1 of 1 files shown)


pwnall:
The `QUIET` issue is separate from the PR's title, and its main goal. Please fork it off to a separate PR, if you think it's worth pursuing. While I'm not the sole maintainer here, I'm opposed to removing `QUIET`, because snappy builds just fine without googletest. In fact, that's how we're currently testing it on Travis / AppVeyor.

gdsotirov:
OK, I returned back QUIET in 7f61e75, but I still think it shouldn't be so if target snappy-unittest that is build by default requires it.

pwnall:
snappy-unittest does not currently require googletest. It has its own stubs, which are used in case googletest is not found. As I've mentioned above, the CI builds do not install googletest or gflags. (feel free to look at `.appveyor.yml` and `.travis.yml` to verify this claim)

gdsotirov:
I trust you of course, but adding the variable is something needed for building snappy under Linux. Even I build only snappy library, then trying to install fails, because it tries to build snappy-unittest as well. Perhaps snappy-unittest should not be build by default?

pwnall:
How are you building snappy on Linux?

This is what we run on CI: `cmake .. && cmake --build .`

If you'd like to avoid building the unit tests (even though they should build just fine), `cmake .. -DSNAPPY_BUILD_TESTS=0 && cmake --build .`

gdsotirov:
I simply do the following:

```mkdir build && cd build
cmake -LAH \
-DCMAKE_BUILD_TYPE=Release \
-DCMAKE_CXX_FLAGS="$SLKCFLAGS" \
-DCMAKE_C_FLAGS="$SLKCFLAGS" \
-DCMAKE_INSTALL_PREFIX=/usr \
-DCMAKE_VERBOSE_MAKEFILE=TRUE \
..
make -j4
make DESTDIR=<somedir> install
```

but OK, I'll add `-DSNAPPY_BUILD_TESTS=0` from now on :-)

PR #47, opened 2017-07-21
Fix minimum version requirement for cmake

nehaljwani:
`WINDOWS_EXPORT_ALL_SYMBOLS` was introduced in `CMake 3.4`


nehaljwani:
I signed it!

pwnall:
This breaks the CI, because the Travis trusty images have CMake 3.2. As 3.4 is only required for Windows, and there are non-Windows deployments that have CMake < 3.4, I'd rather not apply this change.

If you think the error message that you get with CMake < 3.4 on Windows is not helpful, please consider submitting a PR that adds a more explicit error message if CMake < 3.4 on Windows, in the `if` branch that adds `WINDOWS_EXPORT_ALL_SYMBOLS`.

PR #48, opened 2017-07-25
Build static version of lib with cmake

MaxRis:


MaxRis:
I signed it!

dougalsutherland:
This is similar to the static-lib part of https://github.com/google/snappy/pull/38, but adds the static lib as a separate target (`snappy_static.lib` on Windows, `libsnappy.a` elsewhere).

pwnall:
We decided to use the approach in https://github.com/google/snappy/commit/c756f7f5d90ad7591d4f96bb0da8159634943359 for consistency with other Google projects. (FWIW, my main source of inspiration is googletest.)

If you have some reason that you absolutely need to build both the static and the shared library at the same time, please comment and let me know.

PR #51, opened 2017-08-04
Readd pkgconfig support.

niclasr:
Readd pkgconfig support to support usage of snappy with other buildsystems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,11 +54,6 @@ configure_file(
   "${PROJECT_SOURCE_DIR}/cmake/config.h.in"
   "${PROJECT_BINARY_DIR}/config.h"
 )
-configure_file(
-  "${PROJECT_SOURCE_DIR}/cmake/snappy.pc.in"
-  "${PROJECT_BINARY_DIR}/snappy.pc"
-  @ONLY
-)
 
 # We don't want to define HAVE_ macros in public headers. Instead, we use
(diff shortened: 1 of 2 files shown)


niclasr:
I signed it!

cmumford:
@googlebot rescan

ffontaine:
`snappy.pc.in` should be updated with the following line so applications statically linking with snappy (such as wireshark) could retrieve this dependency: 
`Libs.private: -lstdc++`

niclasr:
Well @ffontaine libstdc++ is the c++ runtime used by gcc, if another compiler like clang is used libc++ is used instead so then it must be -lc++. so hardcoding it is not a good idea. wireshark does not use pkgconfig to find snappy from what I have seen, it uses find_package(Snappy). wireshark obviously does not use the c++ compiler when linking to snappy and thereby missing the c++ runtime. Snappy can make the software that depend on it via find_package to use c++ for linking with the static library by setting the property IMPORTED_LINK_INTERFACE_LANGUAGES to CXX on the snappy target and using the keyword EXPORT_LINK_INTERFACE_LIBRARIES in the install(EXPORTS ...) command. I think that would be something to add.

pwnall:
We will not add pkg-config support for now.

We don't want to get into supporting individual package managers, because the C++ ecosystem hasn't converged into a single solution, and there are quite a few equally reasonable candidates. We can't justify the investment into supporting all the package managers, or into supporting a couple of package managers and trying to rationalize where we drew the line.

PR #68, opened 2018-06-01
Update "latest version" on web site to 1.1.7

ryandesign:
The web site says:

> The latest release at the time of writing is 1.1.3.

but that hasn't been true for a year and a half.

This PR updates the version and the download link to 1.1.7.

Since snappy no longer offers "release" downloads, since they're no longer necessary now that snappy has switched from autotools to cmake, the new link is an "archive" download.

The "master" links are also switched to "archive" downloads to match.

I am not very familiar with the automatic GitHub Pages generator that was originally used to create this site. I hope it is correct to update both the params.json and index.html files, as I've done here.

My CLA is on file.


ryandesign:
Do you need anything else from me in order to be able to merge this PR?

cmumford:
@googlebot rescan

cmumford:
Thanks for the PR @ryandesign. There's an upcoming change that removes this line entirely. It's not important to duplicate this version and it'll just get out-of-date.

PR #73, opened 2018-08-27
fix cmake build error

LYKZZzz:
I want to install libsnappy on linux mint 19. I already install GTest from github, but make complained undefined reference:
```
[100%] Linking CXX executable snappy_unittest
CMakeFiles/snappy_unittest.dir/snappy_unittest.cc.o: In function `snappy::Snappy_ZeroOffsetCopy_Test::TestBody()':
snappy_unittest.cc:(.text+0x680b): undefined reference to `testing::Message::Message()'
```

It seems like CMakeList not link0 to libgtest.  Fix it by adding ```target_link_libraries(snappy_unittest gtest)``` in CMakelists.txt


cmumford:
@googlebot rescan

pwnall:
If you're installing snappy for use with other projects, you should use the following CMake flags:

```bash
cmake  -DSNAPPY_BUILD_TESTS=0 -DCMAKE_BUILD_TYPE=Release ..
```

@cmumford WDYT about updating our README files across snappy, leveldb and crc32c? I'm now mildly concerned that folks are following our instructions to the letter, and burning power running un-optimized versions of our libraries.

cmumford:
@pwnall I'm OK with changing the README's

xkszltl:
May I ask why the unit test should be disabled?
In my use case I build the latest snappy and pack it into deb/rpm for use, in both docker and host environment.
I'd like to have a way to quickly verify the lib is working by its own, before installing and providing it to downstream dependencies

ShraddhaDevaiya:
I was finding about the link0 command in Linux but  I didn't get much and I found this conversation.
So can you please help to know about what is link0 actually, If possible ? because I need to use it somewhere else.

pwnall:

PR #74, opened 2018-10-03
Add R snappier package

HuwCampbell:
I've added my package to the list for R, as
- it is a complete package on CRAN;
- it doesn't require any extra linking or building and;
- it doesn't suffer from runtime exceptions like the R-snappy package does when given specific inputs.


HuwCampbell:
I signed it!

cmumford:
@googlebot rescan

cmumford:
@googlebot rescan

PR #75, opened 2018-12-14
Update README to use HTTPS instead of HTTP.

Maikuolan:
HTTPS is currently available for all the HTTP links included in the README.
As such, using HTTPS instead of HTTP for those links may be preferable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ Usage
 
 Note that Snappy, both the implementation and the main interface,
 is written in C++. However, several third-party bindings to other languages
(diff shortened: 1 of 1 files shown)


Maikuolan:
@pwnall No pressure at all, but just asking for the sake of clarity: Is this repository regarded as still being currently maintained/active, or it is dead/EoL/etc?

cmumford:
@googlebot rescan

Devildancing666 (approved):

PR #78, opened 2019-06-28
Add libFuzzer harness and cmake option to build it

bshastry:
Hello,

I noticed that snappy is a security critical dependency of chromium and decided to create a fuzzer for it.

This PR
  - adds a libFuzzer style harness for fuzzing snappy's compressor-decompressor
  - adds a CMake option (off by default) for build the said harness

Once this PR is reviewed and approved, I plan to create a PR to oss-fuzz to upstream this fuzzer.

CC @dor1s @inferno-chromium @kcc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,8 +72,11 @@ before_script:
 - mkdir -p build && cd build
 - cmake .. -G Ninja -DCMAKE_BUILD_TYPE=$BUILD_TYPE
     -DSNAPPY_REQUIRE_${CPU_LEVEL}=ON
+    -DSNAPPY_BUILD_FUZZERS=ON
 - cmake --build .
 - cd ..
 
 script:
 - build/snappy_unittest
+- build/snappy_compress_uncompress_fuzzer -runs=1000 -close_fd_mask=3
+- build/snappy_uncompress_fuzzer -runs=1000 -close_fd_mask=3
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,9 +70,16 @@ install:
 
 before_script:
 - mkdir -p build && cd build
-- cmake .. -G Ninja -DCMAKE_BUILD_TYPE=$BUILD_TYPE
+- if [ "$CXX" = "clang++-8"; then
+    # Build UBSan binary
+    export CXXFLAGS="-fsanitize=undefined -fsanitize-coverage=bb,trace-pc-guard";
+    cmake .. -G Ninja -DCMAKE_BUILD_TYPE=$BUILD_TYPE
     -DSNAPPY_REQUIRE_${CPU_LEVEL}=ON
-    -DSNAPPY_BUILD_FUZZERS=ON
+    -DSNAPPY_BUILD_FUZZERS=ON;
+  else
+    cmake .. -G Ninja -DCMAKE_BUILD_TYPE=$BUILD_TYPE
+    -DSNAPPY_REQUIRE_${CPU_LEVEL}=ON;
+  fi
 - cmake --build .
 - cd ..
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,8 @@ option(BUILD_SHARED_LIBS "Build shared libraries(DLLs)." OFF)
 
 option(SNAPPY_BUILD_TESTS "Build Snappy's own tests." ON)
 
+option(SNAPPY_BUILD_FUZZERS "Build Snappy's fuzzers." OFF)
+
 option(SNAPPY_REQUIRE_AVX "Target processors with AVX support." OFF)
 
 option(SNAPPY_REQUIRE_AVX2 "Target processors with AVX2 support." OFF)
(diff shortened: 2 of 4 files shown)


Dor1s:
+1 to adding this and enabling fuzzing on OSS-Fuzz. Thanks @bshastry !

bshastry:
Hello,

A gentle reminder that this PR is up for review :-)

Thank you

Dor1s:
Hi @pwnall Victor, could you please take a look at this PR? It's important as it should allow us to enable continuous fuzzing for snappy on OSS-Fuzz. Thank you.

pwnall:
Sorry for the delayed response, I just got back from a 2 week vacation. I’ll try to review this PR this week.

pwnall:
High-level comments:

1) This doesn’t seem to match Google’s C++ style guide. I’m OK with accepting the PR as-is, and re-formatting the code myself, if you’re OK with the fact that your name will disappear from most of the git blame output. If you’d rather reformat yourself, here is the style guide: https://google.github.io/styleguide/cppguide.html

clang-format should be able to do most of this for you, if configured correctly.

2) We’ll want at least one more fuzzer, which feeds random data straight into the decompress function. So, the CMake config should say “fuzzers”, instead of “fuzzer”.

3) There is no CI coverage for the fuzzing code. At the very least, `.travis.yml` would need to set `-D SNAPPY_BUILD_FUZZER` in the CMake invocation, so we know that the code compiles. Ideally, CI should run the fuzzer for a small amount of time, so can avoid introducing “easy” bugs.

4) Is this at all related to the snappy code fuzzing in the Chromium repo?

https://cs.chromium.org/chromium/src/testing/libfuzzer/fuzzers/snappy_compress_fuzzer.cc
https://cs.chromium.org/chromium/src/testing/libfuzzer/fuzzers/snappy_uncompress_fuzzer.cc

bshastry:
Hello @pwnall 

Thank you for the comments, I'll get back to you once I've addressed all of them.

I wasn't aware of a similar fuzzer in chromium's source tree. Is that being continuously fuzzer?

Since there seems to be some overlap, would you @Dor1s @pwnall be still interested in having this upstreamed to oss fuzz?

Dor1s:
> Since there seems to be some overlap, would you @Dor1s @pwnall be still interested in having this upstreamed to oss fuzz?

I think it's still useful (we have other projects that are being fuzzed at both places), unless @pwnall is against it :)

One of the advantages is that OSS-Fuzz is testing the tip-of-tree, while Chromium uses a fixed revision (e.g. current snappy's revision is ~6 months old).

One obvious concern might be around duplicated bugs, but I don't think we get many snappy bugs, and I'll actually be happy to see a duplicate rather than continue seeing none :)

It's actually recommended to upstream the fuzz targets, so that the same fuzzers can be used at both infrastructures. Once it's done, I can copy corpus from Chromium's ClusterFuzz instance over to OSS-Fuzz to give it some head start.

bshastry:
Hi @pwnall 

I tried to conform this PR to Google's style guide as much as possible. Still, if there is anything amiss, please go ahead and adapt it as you deem fit. It is fine if my name disappears from git history.

I also added an uncompress only fuzzer, renamed the cmake build option to "fuzzers" instead of "fuzzer" and added the said option to the travis config.

The travis build itself should succeed, however the running itself is of no value as things stand. This is because, the fuzzers need to be built with coverage instrumentation and sanitizers i.e., additional compiler/linker flags need to be passed to the project's build configuration. Moreover, there is a hard dependency on the clang toolchain (as far as I understand, gcc does not yet support ubsan and msan).

Locally, I could get the snappy library built and the fuzzer linked against it and running with the following minimal configuration.
```
$ CC=clang CXX=clang++ CXXFLAGS="-fsanitize=undefined -fsanitize-coverage=bb,trace-pc-guard" cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Debug -DSNAPPY_BUILD_FUZZERS=ON
$ cmake --build .
$ ./snappy_compress_uncompress_fuzzer -runs=1000 -close_fd_mask=3
```

There is another issue to be addressed. Presently the uncompress only fuzzer crashes on null input because this assertion fails. I'm not sure if this is expected or my test harness is buggy. 

https://github.com/google/snappy/pull/78/files#diff-89ca0e55d8c621312263156bc5463918R62

Feedback welcome. Thank you for reviewing.

bshastry:
@pwnall I updated the travis configuration to build fuzzer binaries only when the underlying compiler is clang. Moreover, the travis fuzzer run is restricted to 1000 runs for the UBSan binary. Extending this setup for ASan would require additional changes. Please let me know what you think.

bshastry:
Hello everyone, I would love to get some feedback on this :-)

inferno-chromium:
@pwnall - i also prefer this in oss-fuzz as it has several advantages, like fuzzing on tip-of-tree trunk. bugs being visible to community/downstream consumers once fixed, etc. Can you please help to review.

bshastry:
Hello everyone, still waiting for feedback on this so I can upstream the fuzzer :-)

inferno-chromium:
Sent email to Victor, sorry for delay.

pwnall:
Sorry for the delay in getting to this.

[The original PR wasn't building on Travis](https://travis-ci.org/google/snappy/builds/557826311), but GitHub didn't show the red status. When I took a first pass, I noticed the build failure, but forgot to flag it here. Sorry!

In the interest of time, I made the needed changes and merged the code. The original commits are still in the Git history, so @bshastry gets the credit.

Dor1s:
Woohooo! Thanks, @pwnall !

bshastry:
Thank you @pwnall . I noticed a small issue cmake file that may be an issue with oss-fuzz

https://github.com/google/snappy/blob/5a57d32566da4172f8178f6502dcbe05f7af0f68/CMakeLists.txt#L123-L129

The problem is that the present merge only supports ASan. However, in the oss-fuzz environment, builds for (ideally) all three sanitizers (address, undefined, memory) will be triggered, and the ubsan and memory sanitizer builds will likely fail because one sanitizer (ASan in this case) cannot be combined with other sanitizers?

IIUC, we would need to make the cmake file agnostic of the sanitizer being used and pass the flags in the travis definition.

CC @Dor1s @inferno-chromium

pwnall:
I think you can pass other sanitizers on the command line. Have you tried `export CXXFLAGS="-fsanitize=undefined"` with the current CMake file?  I was hoping it'd work because I'm always appending to existing flags.

I tried UBsan and it currently fails on my machine, complaining about unaligned memory accesses. This can probably be fixed by switching UNALIGNED_LOAD macros to use memcpy, and rely on the optimizer to do the right thing. All modern compilers are doing the right thing: https://godbolt.org/z/QgXaYl

The libFuzzer docs say that MSan support is experimental, so I didn't enable it. https://llvm.org/docs/LibFuzzer.html#id23

pwnall:
Oh the unaligned load situation might be more complicated. https://github.com/abseil/abseil-cpp/blob/master/absl/base/internal/unaligned_access.h makes me think we'll need special macros for sanitizers.

bshastry:
> I think you can pass other sanitizers on the command line. Have you tried export CXXFLAGS="-fsanitize=undefined" with the current CMake file? I was hoping it'd work because I'm always appending to existing flags.

No. Perhaps, I should clarify.

- ossfuzz expects that projects can be built with sanitizers individually (i.e., one sanitizer at a time)
- I stand corrected. Only ASan and MSan are mutually exclusive

> The libFuzzer docs say that MSan support is experimental, so I didn't enable it. https://llvm.org/docs/LibFuzzer.html#id23

@Dor1s @inferno-chromium Is MSan support in libfuzzer still experimental?

inferno-chromium:
MSan works and is not experimental. @oliverchang

PR #79, opened 2019-07-12
Add .gitignore to ignore build artifacts

tmm1:
Makes `git status` clean on a fresh clone + build.


pwnall:
Agreed that having a `.gitignore` would be a good idea.

That being said... what version of snappy are you using? `master` uses CMake, and all build artifacts should be isolated to a single directory. The README recommends `build/` as the build directory name, so I think that's all

tmm1:
Looks like I'm on 32d6d7d8a2ef328a2ee1dd40f072e21f4983ebda which was a bit before v1.1.4

tmm1:
Can we atleast add `/build` to `.gitignore` then?

pwnall:
That sounds good. Would you like to update your PR, or should we pursue that separately?

tmm1:
I will close this PR. I can create a new PR but I think it would be simpler for you to just create a .gitignore with the single entry.

PR #83, opened 2019-08-22
Fix a compile error when the HAVE_SYS_TIME_H flag is not set

HenryRLee:
As <sys/time.h> is guarded by HAVE_SYS_TIME_H, if the flag is missing,
gettimeofday will become an undefined identifier.

### DIFF
--- a/snappy-test.cc
+++ b/snappy-test.cc
@@ -101,7 +101,7 @@ void StartBenchmarkTiming() {
   FILETIME dummy;
   CHECK(GetProcessTimes(
       GetCurrentProcess(), &dummy, &dummy, &dummy, &benchmark_start_cpu));
-#else
+#elif defined HAVE_SYS_TIME_H
   gettimeofday(&benchmark_start_real, NULL);
   if (getrusage(RUSAGE_SELF, &benchmark_start_cpu) == -1) {
     perror("getrusage(RUSAGE_SELF)");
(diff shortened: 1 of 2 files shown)


HenryRLee:
Basically the compilation fails when compiling without using CMake, so the HAVE_SYS_TIME_H flag,  which might be missing.

One simple example is:
```
$ g++ *.cc -std=c++14
In file included from snappy-test.cc:42:
./snappy-test.h:202:5: error: use of undeclared identifier 'gettimeofday'
    gettimeofday(&start_, NULL);
    ^
./snappy-test.h:218:5: error: use of undeclared identifier 'gettimeofday'
    gettimeofday(&stop, NULL);
    ^
snappy-test.cc:105:3: error: use of undeclared identifier 'gettimeofday'
  gettimeofday(&benchmark_start_real, NULL);
  ^
snappy-test.cc:146:3: error: use of undeclared identifier 'gettimeofday'
  gettimeofday(&benchmark_stop_real, NULL);
  ^
4 errors generated.
In file included from snappy_unittest.cc:40:
./snappy-test.h:202:5: error: use of undeclared identifier 'gettimeofday'
    gettimeofday(&start_, NULL);
    ^
./snappy-test.h:218:5: error: use of undeclared identifier 'gettimeofday'
    gettimeofday(&stop, NULL);
    ^
2 errors generated.
```

HenryRLee:
Hi @pwnall, any chance to take a review on this PR?

PR #85, opened 2019-10-17
Allow build with different standard if lib used as a subproject

bitomaxsp:


pwnall:

PR #67, opened 2018-03-25
In gcc __powerpc64__ not __ppc64__ defines the PPC64 architecture

grooverdan:
Corrects 18488d6212331fee647ecfded85353ab3ad91de8 and still maintains
clang compatibility (like the #27 originally).

Compiler tests:

$ uname -m
ppc64le
$  cat  /tmp/x.c

$ gcc -c /tmp/x.c
/tmp/x.c:3:2: warning: #warning __powerpc64__ exists [-Wcpp]
 #warning __powerpc64__ exists
  ^~~~~~~
/tmp/x.c:11:2: error: #error __ppc64__ not defined
 #error __ppc64__ not defined
  ^~~~~
$ clang  /tmp/x.c
/tmp/x.c:3:2: warning: __powerpc64__ exists [-W#warnings]
 ^
/tmp/x.c:9:2: warning: __ppc64__ exists [-W#warnings]
 ^
2 warnings generated.

I should be on the CLA list already.

PR #91, opened 2020-02-20
Add more definitions to detect PPC64

Flamefire:
GCC uses `__PPC64__` not the lower-case version. Not sure about the origin of the lower-case version but I verified the existence of both `__PPC64__` and `__powerpc64__` on our system with GCC 8.3.0.

Note that this is an alternative version of https://github.com/google/snappy/pull/67.

IMO this is more suitable because by the usage of `ARCH_PPC` I assume that a 64 bit version is what is wanted, not a general "is PowerPC", see e.g. here: https://github.com/google/snappy/blob/3fcbc47f9976c880d1e915b1c7cf17633e70e772/snappy-internal.h#L92

### DIFF
--- a/snappy-stubs-internal.h
+++ b/snappy-stubs-internal.h
@@ -72,7 +72,7 @@
 // Enable 64-bit optimized versions of some routines.
 #define ARCH_K8 1
 
-#elif defined(__ppc64__)
+#elif defined(__ppc64__) || defined(__PPC64__) || defined(__powerpc64__)
 
 #define ARCH_PPC 1
 


grooverdan:
Works for me.

FYI, architecture specification: http://openpowerfoundation.org/wp-content/uploads/resources/leabi/content/dbdoclet.50655243_75216.html

pwnall:
Based on the documentation here, I plan to switch to __PPC64__ exclusively. My planned change is at https://github.com/pwnall/snappy/commit/a9955e5d43b6804cacfa56d2d85bcea272ee10f2

Based on the comment above, I think this approach would work for you. Please let me know if you see any problems with it. I'll leave this PR open until something that fixes PPC64 detection lands on master.

Flamefire:
Did you verify other compilers that work on ppc? Like (maybe Clang and icc)? I see no downside to using the 3 macros especially as e.g. the lower-case version seems to have worked for someone.

pwnall:
I landed something in master.

I don't have access to a PPC machine, so the code is 100% based on docs. I did a very superficial attempt of getting PPC up on Travis CI at https://github.com/pwnall/snappy/tree/travis-arch and I got stuck on https://bugs.llvm.org/show_bug.cgi?id=45947

I won't have time to push this line of work forward. If anyone here is interested, please let me know if you make progress!

PR #98, opened 2020-04-08
Minimal changes to add reasonable support for POWER architecture 

dmiller423:
"Minimal changes to add reasonable support for POWER architecture detection,  for use in IncrementalCopy().  Power9 has true unaligned loads and stores, Power8 uses the equivalent of the SSE2 code."

I hope you will consider pulling,  the performance for this is roughly similar to the gain from the SSE2 code for Power8 and a bit higher for Power9.

I may have other optimizations as well, If you are interested in support for this architecture.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,30 +69,12 @@ check_library_exists(z zlibVersion "" HAVE_LIBZ)
 check_library_exists(lzo2 lzo1x_1_15_compress "" HAVE_LIBLZO2)
 
 include(CheckCXXCompilerFlag)
-CHECK_CXX_COMPILER_FLAG("/arch:AVX" HAVE_VISUAL_STUDIO_ARCH_AVX)
-CHECK_CXX_COMPILER_FLAG("/arch:AVX2" HAVE_VISUAL_STUDIO_ARCH_AVX2)
-CHECK_CXX_COMPILER_FLAG("-mavx" HAVE_CLANG_MAVX)
-CHECK_CXX_COMPILER_FLAG("-mbmi2" HAVE_CLANG_MBMI2)
-if(SNAPPY_REQUIRE_AVX2)
-  if(HAVE_VISUAL_STUDIO_ARCH_AVX2)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /arch:AVX2")
-  endif(HAVE_VISUAL_STUDIO_ARCH_AVX2)
-  if(HAVE_CLANG_MAVX)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mavx")
-  endif(HAVE_CLANG_MAVX)
-  if(HAVE_CLANG_MBMI2)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mbmi2")
-  endif(HAVE_CLANG_MBMI2)
-elseif (SNAPPY_REQUIRE_AVX)
-  if(HAVE_VISUAL_STUDIO_ARCH_AVX)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /arch:AVX")
-  endif(HAVE_VISUAL_STUDIO_ARCH_AVX)
-  if(HAVE_CLANG_MAVX)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mavx")
-  endif(HAVE_CLANG_MAVX)
-endif(SNAPPY_REQUIRE_AVX2)
-
 include(CheckCXXSourceCompiles)
+
+include(CheckSymbolExists)
+check_symbol_exists("mmap" "sys/mman.h" HAVE_FUNC_MMAP)
+check_symbol_exists("sysconf" "unistd.h" HAVE_FUNC_SYSCONF)
+
 check_cxx_source_compiles("
 int main() {
   return __builtin_expect(0, 1);
@@ -103,7 +85,33 @@ int main() {
   return __builtin_ctzll(0);
 }" HAVE_BUILTIN_CTZ)
 
-check_cxx_source_compiles("
+
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "i686.*|i386.*|x86.*|amd64.*|AMD64.*|x86_64.*")
+
+  CHECK_CXX_COMPILER_FLAG("/arch:AVX" HAVE_VISUAL_STUDIO_ARCH_AVX)
+  CHECK_CXX_COMPILER_FLAG("/arch:AVX2" HAVE_VISUAL_STUDIO_ARCH_AVX2)
+  CHECK_CXX_COMPILER_FLAG("-mavx" HAVE_CLANG_MAVX)
+  CHECK_CXX_COMPILER_FLAG("-mbmi2" HAVE_CLANG_MBMI2)
+  if(SNAPPY_REQUIRE_AVX2)
+    if(HAVE_VISUAL_STUDIO_ARCH_AVX2)
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /arch:AVX2")
+    endif(HAVE_VISUAL_STUDIO_ARCH_AVX2)
+    if(HAVE_CLANG_MAVX)
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mavx")
+    endif(HAVE_CLANG_MAVX)
+    if(HAVE_CLANG_MBMI2)
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mbmi2")
+    endif(HAVE_CLANG_MBMI2)
+  elseif (SNAPPY_REQUIRE_AVX)
+    if(HAVE_VISUAL_STUDIO_ARCH_AVX)
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /arch:AVX")
+    endif(HAVE_VISUAL_STUDIO_ARCH_AVX)
+    if(HAVE_CLANG_MAVX)
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mavx")
+    endif(HAVE_CLANG_MAVX)
+  endif(SNAPPY_REQUIRE_AVX2)
+
+  check_cxx_source_compiles("
 #include <tmmintrin.h>
 
 int main() {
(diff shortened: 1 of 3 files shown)


dmiller423:
@googlebot I signed it!

dmiller423:
It looks like the travis-ci failure for a specific Xcode build is a missing file,  has nothing to do with the pull request.

CMakeLists.txt
```diff
@@ -103,7 +85,33 @@ int main() {
   return __builtin_ctzll(0);
 }" HAVE_BUILTIN_CTZ)
 
-check_cxx_source_compiles("
+
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "i686.*|i386.*|x86.*|amd64.*|AMD64.*|x86_64.*")
```
grooverdan, line 89: only x86_64 and i386 are needed (uname -p)
ref: https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html#variable:CMAKE_HOST_SYSTEM_PROCESSOR

CMakeLists.txt
```diff
@@ -115,15 +123,43 @@ int main() {
...
+  check_cxx_source_compiles("
 #include <immintrin.h>
 int main() {
   return _bzhi_u32(0, 1);
 }" SNAPPY_HAVE_BMI2)
 
-include(CheckSymbolExists)
-check_symbol_exists("mmap" "sys/mman.h" HAVE_FUNC_MMAP)
-check_symbol_exists("sysconf" "unistd.h" HAVE_FUNC_SYSCONF)
+elseif (CMAKE_SYSTEM_PROCESSOR MATCHES "PPC64*|ppc64*|powerpc64*")
```
grooverdan, line 132: MATCHES "ppc64" is sufficient for LE and BE.

CMakeLists.txt
```diff
@@ -115,15 +123,43 @@ int main() {
...
+
+# Ensure -mcpu=native , many distros still default to power8
+
+  set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -mcpu=native")
```
grooverdan, line 136: I'd only add mcpu=native is another arch specified compile flag isn't used - like https://github.com/facebook/rocksdb/pull/6713/

snappy.cc
```diff
@@ -252,6 +279,42 @@ inline char* IncrementalCopy(const char* src, char* op, char* const op_limit,
       if (SNAPPY_PREDICT_TRUE(op >= op_limit)) return op_limit;
     }
     return IncrementalCopySlow(src, op, op_limit);
+
+#elif SNAPPY_HAVE_PWR9
```
grooverdan, line 283: could/should `#ifdef defined(_ARCH_PWR9)` be used directly?

snappy.cc
```diff
@@ -252,6 +279,42 @@ inline char* IncrementalCopy(const char* src, char* op, char* const op_limit,
...
+
+      char* op_end = std::min(op_limit, buf_limit - 15);
+      while (uint64_t(op_limit-op) >= ns) {
+          vec_xst_len(vu8, (uint8_t*)op, ns);
+          op += ns;
+      }
+      if (SNAPPY_PREDICT_TRUE(op >= op_limit)) return op_limit;
+      return IncrementalCopySlow(src, op, op_limit);
+
+#elif SNAPPY_HAVE_VSX
```
grooverdan, line 299: same for `#elif  defined(__VSX__)`

CMakeLists.txt
dmiller423: Yes it is probably overkill, and i'm more than happy to change/minimize it.  We had some problems with two builds at one point for a project I worked on (reicast emulator) and did not get the standard results.  

For reference current (7.0) msys2/mingw
MINGW64: ``CMAKE_SYSTEM_PROCESSOR: "AMD64"``
MINGW32: ``CMAKE_SYSTEM_PROCESSOR: "AMD64"``  ( yes this is it's output )

$ rm -rf ./.* ./* ; file /mingw32/bin/gcc.exe ;  uname -a ; cmake.exe -G "MinGW Makefiles" ..
/mingw32/bin/gcc.exe: PE32 executable (console) Intel 80386 (stripped to external PDB), for MS Windows
MINGW32_NT-10.0-19041 dev-station 3.0.7-338.x86_64 2019-07-11 10:58 UTC x86_64 Msys

CMAKE_SYSTEM_PROCESSOR: "AMD64"
CMAKE_SIZEOF_VOID_P: "4"
CMAKE_C_COMPILER_ID: "GNU"


The other system didn't match x86_64 || i386 was a modified unix (bsd) toolchain.   Which brings up the possibility for a cmake toolchain file to specify anything It wants via   set(CMAKE_SYSTEM_PROCESSOR x64)

Anyhow I will limit it to the normal out

CMakeLists.txt
dmiller423: This one is a bit more complex,  most of the distros currently are compiled for power8 so the default build flags are for -mcpu=power8.   While we could dynamically check via elf aux inputs or reading special regs if we had privs to do so,  the code shown on the facebook page is still all done at compile time and doesn't seem to have any benefit compared to using -mcpu=native so I'm not sure what the argument is for.   
 
It's a tradeoff whether you want the code added to check dynamically and add a branch to change codepath at runtime,    keep -mcpu=native or another method to check at compile time what cpu it's being built on,  or do nothing and assume the default flags are correct or the user has modified the CLAGS or CMAKE_C_FLAGS.

snappy.cc
dmiller423: I am just copying what was done already for SSE,  the same could be done via SSE preprocessor flags yet it uses it's own.  I am simply trying to keep the style as I found it.

CMakeLists.txt
grooverdan: Agree its a few tough tradeoffs here and its up to what the maintainers accept. My main concern if a distro ends up with a p9 build machine but still supports P8. `-mcpu=native` ends up after `-mcpu=power8` and even I'm not sure what happens then.

Alternatives, [__builtin_cpu_supports](https://gcc.gnu.org/onlinedocs/gcc/Basic-PowerPC-Built-in-Functions-Available-on-all-Configurations.html#index-_005f_005fbuiltin_005fcpu_005fsupports) or [\_\_target\_\_](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html) attributes on the function.

dmiller423:
If it's built on power9 and the resulting elf run on power8, the result
would be an exception: illegal instruction.  This is what package
management is for, or building for the system you intend to run it on.

This case is only an issue if a binary is distributed incorrectly.

astubbs:
Is there an active issue tracker for snappy? the old google code issue links don't work, and GH issues is disabled for the repo. Can we turn on the GH issues feature?

abebeos:
> 
> 
> Is there an active issue tracker for snappy? the old google code issue links don't work, and GH issues is disabled for the repo. Can we turn on the GH issues feature?

It looks that for now, you can use the group: https://groups.google.com/g/snappy-compression to (kind of) file issues.

snappy.cc
abebeos: Would say that he normalized version is used correclty (`SNAPPY_HAVE_VSX`).

@dmiller423 , remember to flag this as "solved"

snappy.cc
abebeos: same, nomalized version should be used. @dmiller423 , should be marked as "solved".

snappy.cc
```diff
@@ -721,8 +784,8 @@ static inline uint32 ExtractLowBytes(uint32 v, int n) {
   return _bzhi_u32(v, 8 * n);
 #else
   // This needs to be wider than uint32 otherwise `mask << 32` will be
-  // undefined.
-  uint64 mask = 0xffffffff;
+  // undefined. 
+  uint64 mask = ~0;
   return v & ~(mask << (8 * n));
```
abebeos, line 789: @dmiller423 You are changing here existent code, which introduces review effort. Are you absolutely(!) sure that this is necessary? If so, the add a comment line, possibly with a link to high-quality documentation. Otherwise just reverse the change.

abebeos:
@dmiller423 , can you please provide a bullet-list stating what is the absolute minimum support you need here for PowerPC? (asking just for the case that this PR does not get merged).

snappy.cc
```diff
@@ -252,6 +279,42 @@ inline char* IncrementalCopy(const char* src, char* op, char* const op_limit,
...
+              memcpy((uint8_t*)&vu8+(ix*pattern_size), src, pattern_size);	// vu|=vu<<pattern_size
+
+          char* op_end = std::min(op_limit, buf_limit - 15);
+          while (op < op_end) {
+              vec_vsx_st(vu8, 0, (uint8_t*)op);
+              op += ns;
+          }
+          if (SNAPPY_PREDICT_TRUE(op >= op_limit)) return op_limit;
+      }
+      return IncrementalCopySlow(src, op, op_limit);
```
abebeos, line 316: are those two sections (SNAPPY_HAVE_PWR9, SNAPPY_HAVE_VSX) the main speed-enhancement code?

pwnall:
Hey folks,

In the hope of avoiding wasted work, we're putting together some guidance around what contributions are more / less likely to be accepted.

POWER is not a focus for us at the moment, so the most we might accept is straightforward changes that don't introduce significant extra complexity. This PR isolates the POWER-related code fairly well, but adds a bunch of complexity to the CMakeFile.

I would advise waiting for us to post guidance before investing more effort into getting something merged in this repository.

To be clear, folks are always welcome to host forks with support for other architectures.

snappy.cc
```diff
@@ -721,8 +784,8 @@ static inline uint32 ExtractLowBytes(uint32 v, int n) {
   return _bzhi_u32(v, 8 * n);
 #else
   // This needs to be wider than uint32 otherwise `mask << 32` will be
-  // undefined.
-  uint64 mask = 0xffffffff;
+  // undefined. 
+  uint64 mask = ~0;
```
abebeos, line 788: (commenting the 2nd time here, comment disappeared)

are you absolutely(!) sure to change "0xffffffff" to  "~0" here? This looks like modifying the existent code, which introduces additional review effort. If confident, provide a comment (possibly with a link to relevant docu).

abebeos:
y folks,
> 
> In the hope of avoiding wasted work, we're putting together some guidance around what contributions are more / less likely to be accepted.
> 
> POWER is not a focus for us at the moment, so the most we might accept is straightforward changes that don't introduce significant extra complexity. This PR isolates the POWER-related code fairly well, but adds a bunch of complexity to the CMakeFile.
> 
> I would advise waiting for us to post guidance before investing more effort into getting something merged in this repository.

If you look at my review comments, you'll notice that I more or less follow "standard-procedure" (simple, decoupled, minimum review effort, ...) here. I would actually immediately reject this PR as a maintainer, and ask for 2 new ones:

* one PR with the general PowerPC additions for cmake (rule: "do not touch existent constructs")
* one PR with the specific PowerPC code (which naturally "does not touch existent constructs of other architectures")

> To be clear, folks are always welcome to host forks with support for other architectures.

Adding a new architecture is actually is a good quality-test (and an opportunity to increase the quality).

abebeos:
@dmiller423, it looks preferable to close this PR.

I've provided the first PR (#124) which contains the minimum cmake related changes. You'll possibly notice that the review effort is near zero.

I'll provide a similar 2nd patch with your `snappy.cc` changes, again normalized/simplified in order to simplify review and keep the code 100% non-intrusive.

(please feel free to repost the patches if you prefer that they appear in your contribution histories. I purge my history anyways every few months).

abebeos:
This PR is superseded by  #126/#124.

@pwnall , can be closed.

abebeos:
@dmiller423 , you should close this PR yourself, as it is:

* outdated
* seemingly non-functional (speed-tests have shown speed loss)

PR #101, opened 2020-05-16
travis: bump linux gcc to 10

grooverdan:
brew doesn't have gcc-10 yet, but linux does.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,8 +41,8 @@ addons:
     packages:
     - clang-10
     - cmake
-    - gcc-9
-    - g++-9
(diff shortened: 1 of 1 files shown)


pwnall:
Is it OK if we wait for this until https://github.com/Homebrew/homebrew-core/pull/54364 gets merged? I would strongly prefer not to have to think about two GCC versions on CI. It seems like there's a lot of activity on the PR, so I'm hoping it would wrap up in a week or so.

pwnall:
Sorry I didn't get back to this earlier! We're on GCC 10 everywhere now.

PR #102, opened 2020-05-18
Add another C# implementation

aloneguid:
[IronSnappy](http://groups.google.com/group/snappy-compression) is new but modern, tested, and fully managed implementation.


aloneguid:
@googlebot I signed it

PR #105, opened 2020-10-07
Availability of the immintrin.h should not implicitly disable support…

gruuk:
… of all pre-Haswell CPUs. User needs to explicitly specify AVX2 requirement via SNAPPY_REQUIRE_AVX2 if it was their intention.


gruuk:
@googlebot I signed it!

gruuk:
@googlebot I fixed it.

PR #106, opened 2020-10-13
Fix docs: Snappy compressor works in 64 kB blocks instead of 32 kB now

hwwhww:
Hi! It looks like https://github.com/google/snappy/commit/27a0cc394950ebdad2e8d67322f0862835b10bd9 changed block size from 32 kB to 64 kB many years ago, but the `format_description.txt` doc was not modified.


hwwhww:
@googlebot I signed it!

PR #108, opened 2020-11-22
Fix compilation in clang, unused variable

Hugmeir:
Compilign with clang in OSX gives this:

    snappy.cc:96:18: error: unused variable 'kLiteralOffset' [-Werror,-Wunused-const-variable]

A recent commit replaced the uses of kLiteralOffset,  but it is
still mentioned in comments, so to keep those making some sense,
this commit comments out the variable too.


Hugmeir:
@googlebot I signed it!

pwnall:
Sorry I wasn't able to get to this earlier! In the meantime, this error was fixed by internal repository changes.

PR #112, opened 2020-12-15
Added -Wno-ignored-attribute for clang/gcc.

cmumford:
Compiler warning when unable to inline a template./

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,6 +71,11 @@ else(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
   endif(NOT CMAKE_CXX_FLAGS MATCHES "-Werror")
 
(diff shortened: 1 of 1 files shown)


pwnall:
How about deciding that we'll support `-Wall -Werror` on clang, but only do best-effort support on GCC?

I'd like us to support `-Wall -Werror` in clang w/o exceptions so we can make it easy to integrate Snappy into other Google products, like Chrome. Also, if needed, I think we could tweak the internal config to make sure that `-Wall` works.

cmumford:
I'm fine with that approach. There seem to be several clang's in [CMAKE_CXX_COMPILER_ID](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html), but only one that appears to be GCC ("GNU"), so I can test against that. I'll put together a second PR with that change.

cmumford:
Replacement PR is https://github.com/google/snappy/pull/113

cmumford:
Oops - 113 was bad. Instead going with https://github.com/google/snappy/pull/114

PR #113, opened 2020-12-15
Building with `-Wall` only for clang.

cmumford:
gcc was unable to inline a function call, which caused a build failure due to `-Wall`.

The build error was:

```
../snappy.cc:292:76: error: ignoring attributes on template argument ‘__m128i’ [-Werror=ignored-attributes]
  292 | static inline std::pair<__m128i /* pattern */, __m128i /* reshuffle_mask */>
      |                                                                            ^
../snappy.cc:292:76: error: ignoring attributes on template argument ‘__m128i’ [-Werror=ignored-attributes]
cc1plus: all warnings being treated as errors
```


cmumford:
Superseded by https://github.com/google/snappy/pull/114

PR #114, opened 2020-12-15
Building with `-Werror` only with clang.

cmumford:
gcc was unable to inline a function call, which caused a build
failure due to `-Wall -Werror`.

The build error was:

```
../snappy.cc:292:76: error: ignoring attributes on template argument ‘__m128i’ [-Werror=ignored-attributes]
  292 | static inline std::pair<__m128i /* pattern */, __m128i /* reshuffle_mask */>
      |                                                                            ^
../snappy.cc:292:76: error: ignoring attributes on template argument ‘__m128i’ [-Werror=ignored-attributes]
cc1plus: all warnings being treated as errors
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,8 +67,7 @@ else(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
   endif(NOT CMAKE_CXX_FLAGS MATCHES "-Wextra")
 
   # Use -Werror for clang only.
-  # There are many clang compiler ID's, but only one gcc.
-  if(NOT CMAKE_CXX_COMPILER_ID STREQUAL "GCC")
+  if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     if(NOT CMAKE_CXX_FLAGS MATCHES "-Werror")
       set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
     endif(NOT CMAKE_CXX_FLAGS MATCHES "-Werror")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,10 +66,13 @@ else(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wextra")
   endif(NOT CMAKE_CXX_FLAGS MATCHES "-Wextra")
 
-  # Use -Werror for clang and gcc.
-  if(NOT CMAKE_CXX_FLAGS MATCHES "-Werror")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -66,10 +66,13 @@ else(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wextra")
   endif(NOT CMAKE_CXX_FLAGS MATCHES "-Wextra")
 
-  # Use -Werror for clang and gcc.
-  if(NOT CMAKE_CXX_FLAGS MATCHES "-Werror")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
-  endif(NOT CMAKE_CXX_FLAGS MATCHES "-Werror")
+  # Use -Werror for clang only.
+  # There are many clang compiler ID's, but only one gcc.
+  if(NOT CMAKE_CXX_COMPILER_ID STREQUAL "GCC")
```
pwnall, line 71: This approach is fine, because we only support Clang, GCC, MSVC.

An alternative to consider is `if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")`

CMakeLists.txt
```diff
@@ -66,10 +66,13 @@ else(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
...
+    if(NOT CMAKE_CXX_FLAGS MATCHES "-Werror")
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
+    endif(NOT CMAKE_CXX_FLAGS MATCHES "-Werror")
+  endif(CMAKE_CXX_COMPILER_ID STREQUAL "GCC")
```
pwnall, line 75: Please make sure the `endif` clause matches the `if` clause. Should be one of the following, depending on which way you decide to go.

* `endif(NOT CMAKE_CXX_COMPILER_ID STREQUAL "GCC")`
* `endif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")`

pwnall (changes requested):
Thank you for putting this together! Here are a couple of comments. You can either apply them here or just adopt them in the internal repository CL.

CMakeLists.txt
cmumford: Done - thank you.

CMakeLists.txt
cmumford: Fixed - thx.

cmumford: LMK if you that nested if/if would read better as a compound if (... AND ...). The line was getting long.

PR #116, opened 2020-12-15
Fix fuzzing on OSX

cmumford:
Testing to see if the xcode11 image has libclang_rt.fuzzer_osx.a where the xcode12 image does not.


cmumford:
this was mostly an experiment to see if this would work with the older image.

PR #124, opened 2021-03-10
Add PowerPC VSX and PWR9 detection

abebeos:
Based on #98 (@dmiller423).

This PR contains the simplified, cmake-related part of #98.

The patch should be completely non-intrusive (= does not affect in any way other code)


abebeos:
closing, as base patch #98 is non-functional (degrades speed)

PR #129, opened 2021-05-28
remove Disable-RTTI flag for the convenience of derivation

holmes1412:
Compilation flag ``Disable RTTI`` has been added in commit: c98344f6260d24d921e5e04006d4bedb528f404a , which will cause a link error for those who derive class in snappy, like ``snappy::Sink`` and ``snappy::Source``, because **RTTI** is enable by default. 
Our project [Workflow](https://github.com/sogou/workflow) derives snappy for sequence of bytes compressing and finds out ``Disable RTTI`` will cause us some trouble.
Considering ``Disable RTTI``  cannot make it easier to adopt snappy in other projects and cannot introduce other obvious advantages either, I suggest removing this compile flag.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,10 +51,6 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
   string(REGEX REPLACE "/EH[a-z]+" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHs-c-")
   add_definitions(-D_HAS_EXCEPTIONS=0)
-
-  # Disable RTTI.
-  string(REGEX REPLACE "/GR" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /GR-")
 else(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
   # Use -Wall for clang and gcc.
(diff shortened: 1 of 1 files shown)


holmes1412:
@googlebot I signed it!

holmes1412:
Here's a simple demo: ``tmp.cc``

```cpp
#include "snappy-sinksource.h"

class T : public snappy::Sink
{
public:
    virtual const char *Peek(size_t *len)
    {   
        const char *pos;
        return pos;
    }   
    virtual size_t Available() const { return 0; }
    virtual void Sink(size_t n) {}
    virtual void Append(const char *bytes, size_t n) {}
};

int main()
{
    T t;
    return 0;
}
```
Compile command:
```sh
g++ tmp.cc libsnappy.a -std=c++11
```
Error:
```sh
/tmp/cctcsFKq.o:(.rodata._ZTI1T[_ZTI1T]+0x10): undefined reference to `typeinfo for snappy::Sink'
collect2: error: ld returned 1 exit status
``

pwnall:
@holmes1412 Thank you for your contribution!

The project's CMake configuration reflects the way it's used in Google Chrome. This is the only configuration we can maintain ourselves.

To be clear, this doesn't mean your changes are not valid -- I think you'll be able to use Snappy with the build tweaks you posted here just fine. It's just that we can't accept this change in the official repository.

I hope this explanation helps.

anatol:
Here is an Arch Linux bug that is caused by disable-rtti flag https://bugs.archlinux.org/task/72058

@pwnall I wonder if this disable-rtti feature needs to have its own flag. Chrome folks can disable `rtti` if they need, but for the standalone package configuration (like in Linux distros) `rtti` flag should probably be enabled.

PR #132, opened 2021-07-09
Add a switch to use system gtest and benchmark

DanySpin97:
Allow to use the gtest and benchmark libraries from the system.
Use pkg-config to check that the libraries are installed and to
add the correct cflags/link flags.

All the switches are enabled by default on UNIX and set to ON for other platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,6 +90,14 @@ option(SNAPPY_BUILD_TESTS "Build Snappy's own tests." ON)
 
 option(SNAPPY_BUILD_BENCHMARKS "Build Snappy's benchmarks" ON)
 
+if(UNIX)
+  option(SNAPPY_USE_BUNDLED_GTEST "Build test using bundled googletest library" ON)
+  option(SNAPPY_USE_BUNDLED_BENCHMARK_LIB "Build benchmarks using bundled benchmark library" ON)
+else(UNIX)
+  set(SNAPPY_USE_BUNDLED_GTEST ON)
+  set(SNAPPY_USE_BUNDLED_BENCHMARK_LIB ON)
+endif(UNIX)
+
 option(SNAPPY_FUZZING_BUILD "Build Snappy for fuzzing." OFF)
 
 option(SNAPPY_REQUIRE_AVX "Target processors with AVX support." OFF)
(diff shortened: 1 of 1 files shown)


pwnall:

Unfortunately, I don't think the benefits outweigh the complexity introduced by this PR.

Google Test and Google Benchmark are only used to develop this library. They are not needed to build the library for a distribution. So, I see very little benefit in supporting any versions other than the exact versions used on CI.

DanySpin97:
> They are not needed to build the library for a distribution.

They are needed already. Distributions runs test of packages, including snappy. Fedora for example is doing something similar [here](https://src.fedoraproject.org/rpms/snappy/blob/rawhide/f/snappy-thirdparty.patch). Other distributions aren't because they haven't updated to 1.1.9, but eventually most of them will have something similar.

If it is a problem of supporting every version of gtest, a minimum version can be added when using pkg-config.

pwnall:
Thank you for the additional explanations.

> > They are not needed to build the library for a distribution.
> 
> They are needed already. Distributions runs test of packages, including snappy. Fedora for example is doing something similar [here](https://src.fedoraproject.org/rpms/snappy/blob/rawhide/f/snappy-thirdparty.patch). Other distributions aren't because they haven't updated to 1.1.9, but eventually most of them will have something similar.

Google Test and Google Benchmark are not needed to *build* the library.

Distributions that want to run tests using a supported configuration should use git's submodules functionality to pull the Google Test and Google Benchmark versions we use to run tests. Distributions that want to run the test suite using other versions of these dependencies are following an unsupported path, and will need to maintain their own patches. 

> If it is a problem of supporting every version of gtest, a minimum version can be added when using pkg-config.

GoogleTest now follows the Abseil Live at Head philosophy, so it's a bit more difficult to reason about a minimum version.

> EDIT: This patch is needed for both openSUSE (where I am currently updating the package) and for Exherbo (where I was updating the package some months ago).

Thank you for pointing out the cost to the ecosystem!

DanySpin97:
> Distributions that want to run the test suite using other versions of these dependencies are following an unsupported path, and will need to maintain their own patches.

So you are telling distributions to stop using release tarballs for snappy, because tarballs don't contain git's submodules.

anatol:
Thank you @DanySpin97 I pulled your fix to the official Arch Linux package and it works good so far.

@pwnall what are the plans to merge this change to the repo?

pwnall:
@anatol We're not planning to support this configuration.

@DanySpin97 Yup, I can't recommend GitHub's tarballs. I'd always clone the repository instead. Something along the following lines should work.

```bash
git clone https://github.com/google/snappy
cd snappy
git checkout tags/1.1.9
git submodule update --init --recursive
```

DanySpin97:
@pwnall Yes, that's how it can be compiled and tested from git, but that's not how distributions work. They use release tarballs, sometimes even include the tarballs in their control version system. In openSUSE this is even mandatory to package a project.

Can you please confirm that you don't intend to support distributions, (considering also #51) and, if so, put a disclaimer about that in the README? I think that would be fair, but I'd just like it this would have been made more clear.

PR #133, opened 2021-07-22
Simd

JunHe77:


JunHe77:
@googlebot I signed it!

JunHe77:
@googlebot I signed it!

JunHe77:
Rebased to resolve conflict in snappy-internal.h

JunHe77:
@atdt , would you pls help to review? Thanks.

snappy-internal.h
```diff
@@ -36,13 +36,44 @@
 namespace snappy {
 namespace internal {
 
+#if !defined(SNAPPY_HAVE_BMI2)
```
pwnall, line 39: Thank you for realizing we need these #defines and #includes in the header, for the `V128` type alias.

Can you please move this block outside namespaces?

I think that #including C system headers in a C++ namespace is undefined behavior.

snappy-internal.h
```diff
@@ -36,13 +36,44 @@
...
+#include <tmmintrin.h>
 using V128 = __m128i;
-#else
+#elif SNAPPY_HAVE_NEON
+#include <arm_neon.h>
 using V128 = uint8x16_t;
 #endif
 
+#if SNAPPY_HAVE_BMI2
+inline uint32_t zero_highbits_u32(uint32_t v, int n) {
```
pwnall, line 73: Could you please explain the benefit of wrapping `_bzhi_u32` with this function?

Until I learn more, I'm inclined to suggest that we remove the abstraction. It's currently used in exactly one place, and it's only defined when `SNAPPY_HAVE_BMI2`. So, it's effectively a less-known alias for the `_bzhi_u32` intrinsic.

pwnall (changes requested):
@JunHe77 thank you very much for this contribution!

Would you like to address the feedback below, or would you prefer that I take care of it while going through our merge process?

snappy-internal.h
JunHe77: Thanks for pointing this out, will update.

snappy-internal.h
JunHe77: Well received, will update.

JunHe77:
Updated. @pwnall , @atdt , would you pls help to review? Thanks a lot.

pwnall (approved):

pwnall:
@JunHe77 Thank you very much for the revisions!

I'll get this PR through the merge process in our internal repository. When the process completes, the PR will be merged.

PR #135, opened 2021-08-03
decompress: add hint to remove extra AND

JunHe77:
Clang doesn't realize the load with free zero-extension,
and emits another extra 'and xn, xm, 0xff' to calc offset.
With this change ,this extra op is removed, and consistent
1.7% performance uplift is observed.

Change-Id: Ica4617852c4b93eadc6c5c551dc3961ffbadb8f0

### DIFF
--- a/snappy.cc
+++ b/snappy.cc
@@ -1108,6 +1108,15 @@ std::pair<const uint8_t*, ptrdiff_t> DecompressBranchless(
     // ip points just past the tag and we are touching at maximum kSlopBytes
     // in an iteration.
     size_t tag = ip[-1];
+#if defined(__clang__) && defined(__aarch64__)
+    // Workaround for https://bugs.llvm.org/show_bug.cgi?id=51317
+    // when loading 1 byte, clang for aarch64 doesn't realize that it(ldrb)
+    // comes with free zero-extension, so clang generates another
+    // 'and xn, xm, 0xff' before it use that as the offset. This 'and' is
+    // redundant and can be removed by adding this dummy asm, which gives
+    // clang a hint that we're doing the zero-extension at the load.
+    asm("" ::"r"(tag));
+#endif
     do {
       // The throughput is limited by instructions, unrolling the inner loop
       // twice reduces the amount of instructions checking limits and also


atdt (changes requested):
Thanks for this, Jun! Could you add a comment explaining what this is doing, and referencing https://bugs.llvm.org/show_bug.cgi?id=51317?

JunHe77:
Thanks for reviewing this, @atdt . Code comments have been added. Pls check. Thanks again.

atdt:
@pwnall this looks good to me.

snappy.cc
```diff
@@ -1108,6 +1108,19 @@ std::pair<const uint8_t*, ptrdiff_t> DecompressBranchless(
...
+#if defined(__clang__) && defined(__aarch64__)
+    // when loading 1 byte, clang for aarch64 doesn't realize that it(ldrb)
+    // comes with free zero-extension, so clang generates another
+    // 'and xn, xm, 0xff' before it use that as the offset. This 'and' is
+    // redundant and can be removed by adding this dummy asm, which gives
+    // clang a hint that we're doing the zero-extension at the load.
+    // The root cause is the bug in clang, which is reported here:
+    // https://bugs.llvm.org/show_bug.cgi?id=51317
+    // This workaround should be removed once the clang upstream has that bug
+    // fixed. For people who is using clang-12 or lower you may choose to
```
pwnall, line 1120: Are we sure that this will be fixed in clang 13? I think that the LLVM bug is still open.

If yes -- can we replace / amend this comment with `&& __clang_major__ < 13` in the `#if` above?

If not -- should we remove the "This workaround..." paragraph altogether?

This would make it a bit more obvious that we don't need the workaround in Clang 13+.

pwnall (changes requested):
@JunHe77 thank you for this contribution!

Would you like to address the feedback below, or would you prefer that I take care of it while going through our merge process?

snappy.cc
```diff
@@ -1108,6 +1108,19 @@ std::pair<const uint8_t*, ptrdiff_t> DecompressBranchless(
     // ip points just past the tag and we are touching at maximum kSlopBytes
     // in an iteration.
     size_t tag = ip[-1];
+#if defined(__clang__) && defined(__aarch64__)
+    // when loading 1 byte, clang for aarch64 doesn't realize that it(ldrb)
+    // comes with free zero-extension, so clang generates another
+    // 'and xn, xm, 0xff' before it use that as the offset. This 'and' is
+    // redundant and can be removed by adding this dummy asm, which gives
+    // clang a hint that we're doing the zero-extension at the load.
+    // The root cause is the bug in clang, which is reported here:
```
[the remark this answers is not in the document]
JunHe77: LGTM, will update.

snappy.cc
JunHe77: I'm not sure this bug will be fixed in clang 13. Would it be OK to only remove "For people ...", so users know that this is a workaround and will be finally removed?

pwnall: In my opinion we could replace the entire comment block with

```c++
    // Workaround for https://bugs.llvm.org/show_bug.cgi?id=51317
```

Snappy always assumes the latest stable compiler version for full optimization support. For example, we submitted 27ff130ff90f167bb55cd62089e2fe5d678c59b6 and 041c608086d00ceb8285b91f034cfa8b84c83d7c as soon as the latest released versions of clang and GCC could perform the optimizations themselves.

JunHe77: Agree, that looks much clearer. Thanks for that. Will update.

JunHe77:
Updated. @pwnall , @atdt , would you pls help to review? Thanks.

pwnall (approved):

pwnall:
@JunHe77 I'll start getting this PR through our merge process tomorrow. Thank you very much for the revisions!

PR #144, opened 2021-10-27
Re-enable RTTI (needed in order to subclass Source, etc.)

tserong:
Commit c98344f in snappy 1.1.9 disabled RTTI, which means the snappy library no longer exports typeinfo for snappy::Source, snappy::Sink, ..., so users of the library can't subclass these classes anymore.

Here's a trivial reproducer:

```
  #include <snappy-sinksource.h>
  class MySource : snappy::Source {
  public:
    size_t Available() const override { return 0; }
    const char *Peek(size_t *len) override { return NULL; }
    void Skip(size_t n) override {}
  };
  int main(int argc, char **argv) {
    MySource m;
    return 0;
  }
```

Try `g++ -o snappy-test ./snappy-test.cc -lsnappy` with the above and the linker will fail with "undefined reference to `typeinfo for snappy::Source'" if RTTI was disabled in the snappy build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,10 +51,6 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
   string(REGEX REPLACE "/EH[a-z]+" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHs-c-")
   add_definitions(-D_HAS_EXCEPTIONS=0)
-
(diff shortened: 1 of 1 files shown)


pwnall:
Sorry, I am not going to accept this PR.

The current CMake configuration reflects the supported build setup for this library, and what runs in CI. RTTI is disabled in Google Chrome, which is the largest testing vehicle for Snappy.

tserong:
Would you mind bumping the soname version then, as disabling RTTI breaks the ABI? A similar issue was raised against leveldb (https://github.com/google/leveldb/issues/927)

PR #146, opened 2021-11-02
update `google.github.io/snappy/` broken links

hoomanna:
relative links used is `https://google.github.io/snappy/` is broken.


RTLer:
@googlebot I signed it!

PR #147, opened 2021-11-24
add option to enable rtti, set default to current behavior

AnticliMaxtic:
Pull requests #144 #143 #129 have all attempted to remove the disable of rtti, but they did not maintain the current build behavior. This change adds a CMake `option` allowing rtti to be enabled if consumers wish to, but it is disabled by default.


pwnall:

dg0yt:
2025: Still overriding user's cflags, interfering with downstream use.

PR #148, opened 2021-12-01
Fix UBSan error (ptr + offset overflow)

pitrou:
As `i + offset` is promoted to a "negative" size_t,
UBSan would complain when adding the resulting offset to `dst`:
```
/tmp/RtmptDX1SS/file584e37df4e/snappy_ep-prefix/src/snappy_ep/snappy.cc:343:43: runtime error: addition of unsigned offset to 0x6120003c5ec1 overflowed to 0x6120003c5ec0
    #0 0x7f9ebd21769c in snappy::(anonymous namespace)::Copy64BytesWithPatternExtension(char*, unsigned long) /tmp/RtmptDX1SS/file584e37df4e/snappy_ep-prefix/src/snappy_ep/snappy.cc:343:43
    #1 0x7f9ebd21769c in std::__1::pair<unsigned char const*, long> snappy::DecompressBranchless<char*>(unsigned char const*, unsigned char const*, long, char*, long) /tmp/RtmptDX1SS/file584e37df4e/snappy_ep-prefix/src/snappy_ep/snappy.cc:1160:15
```

### DIFF
--- a/snappy.cc
+++ b/snappy.cc
@@ -340,7 +340,7 @@ static inline bool Copy64BytesWithPatternExtension(char* dst, size_t offset) {
   if (SNAPPY_PREDICT_TRUE(offset < 16)) {
     if (SNAPPY_PREDICT_FALSE(offset == 0)) return false;
     // Extend the pattern to the first 16 bytes.
-    for (int i = 0; i < 16; i++) dst[i] = dst[i - offset];
+    for (int i = 0; i < 16; i++) dst[i] = (dst - offset)[i];
     // Find a multiple of pattern >= 16.
     static std::array<uint8_t, 16> pattern_sizes = []() {
       std::array<uint8_t, 16> res;


rvandermeulen:
We also encountered this UBSan error in Firefox' CI when testing updating from 1.1.8 to 1.1.9. I can confirm that this patch results in a clean test run on our end as well.

pitrou:
cc @emkornfield in case you know how we can push this forward. This is breaking the R CI for Apache Arrow.

emkornfield:
I'll see if I can find the right contact.

jonkeane:
I would love to be able to pull out some of the hacks we have in our CI + tests more general in Arrow to work around this. Is there anything we could do to help push this forward?

emkornfield:
Haven't been able to track down the right person for this, will try again after the holiday weekend.

jonkeane:
Thanks for the help!

snappy.cc
```diff
@@ -340,7 +340,7 @@ static inline bool Copy64BytesWithPatternExtension(char* dst, size_t offset) {
   if (SNAPPY_PREDICT_TRUE(offset < 16)) {
     if (SNAPPY_PREDICT_FALSE(offset == 0)) return false;
     // Extend the pattern to the first 16 bytes.
-    for (int i = 0; i < 16; i++) dst[i] = dst[i - offset];
```
emkornfield, line 343: what there s specific reason for this formulation vs pure pointer arithmetic?  `*(dst - offset + i)`

emkornfield:
@pwnall could you merge this?  Or should we ping someone else?

jonkeane:
Thank you!

PR #149, opened 2021-12-30
Replace biconditional logical connective 'iff' with 'if'.

TheVice:
Hello.

If that was not the typos at the comment sections I assume that was using of [biconditional logical connective](https://en.wikipedia.org/wiki/If_and_only_if) that probably not necessary in that cases.

Proposed to replace it with 'if'.

Thank you.

P.S.
There is a place in third-party section but that is a git sub-module so not changed it via this request.

### DIFF
--- a/snappy-sinksource.h
+++ b/snappy-sinksource.h
@@ -117,7 +117,7 @@ class Source {
   virtual size_t Available() const = 0;
 
   // Peek at the next flat region of the source.  Does not reposition
-  // the source.  The returned region is empty iff Available()==0.
+  // the source.  The returned region is empty if Available()==0.
   //
   // Returns a pointer to the beginning of the region and store its
   // length in *len.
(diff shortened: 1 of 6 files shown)


Quuxplusone:
"iff" in this context means "if and only if". "We do X iff Y" is a much stronger statement than merely "We do X if Y." The former implies "Any time you see X, it's because Y." The latter doesn't imply that; it permits that maybe sometimes X happens without Y's having happened first.
Please keep the stronger form, iff (😄) it is in fact correct.

TheVice:
Well, if seen from that side, you may be right and this change should not
be made.
Thank you for your feedback and explanation.

пн, 1 січ. 2024 р. о 06:11 Quuxplusone ***@***.***> пише:

TheVice:
@Quuxplusone well, if seen from that side, you may be right and this change should not
be made.
Thank you for your feedback and explanation.

PR #155, opened 2022-05-23
Add new framing chunk types without checksums

derekbruening:
Adds two new chunk types to the Snappy framing format: compressed data
without a checksum, and uncompressed data without a checksum.  These
types are identical to their existing counterparts except they do not
contain a CRC-32C checksum.  Essentially, this makes including
checksums for each data chunk optional rather than required.

In some use cases, computing the CRC-32C checksums for the data chunks
in the Snappy framing format ends up dominating execution time.
Eliminating the checksums provides massive 2.5x performance
improvements in our uses of Snappy for compressing address trace data
prior to storing to disk.

Existing readers of the Snappy framing format would be expected to
fail up front on an unknown chunk type when encountering the new
types, until updated to handle them, which should be a simple coding
change.

### DIFF
--- a/framing_format.txt
+++ b/framing_format.txt
@@ -1,9 +1,9 @@
 Snappy framing format description
-Last revised: 2013-10-25
+Last revised: 2022-05-23
 
 This format decribes a framing format for Snappy, allowing compressing to
 files or streams that can then more easily be decompressed without having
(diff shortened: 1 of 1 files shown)


derekbruening:
@pwnall PTAL

PR #159, opened 2022-09-26
Fix warnings due to use of __attribute__(always_inline) without inline

jbms:


jbms:
Merged in https://github.com/google/snappy/commit/d644ca87706548b9a8bf1e13d7669c13c2607b7e

PR #162, opened 2023-01-11
Optimize Store32 to generate reliable single str

JunHe77:
Original implementation generates different instruction stream for Store32. In certain flow it is 4 strb ops which is not the optimisim. Use memcpy for little endian to get stable codegen. Observed ~1-2% uplift on N1.

Change-Id: I5d6d960f6708ea49d44dca709f3de2f1f25737da


JunHe77:
Abandon this as it is covered by https://github.com/google/snappy/pull/150

PR #172, opened 2023-10-04
Add missing <functional> include

assignUser:
`std::less` is defined in `<functional>` which is missing from snappy.cc and causes compilation issues with older compilers e.g. [XCode 10 AppleClang](https://github.com/ursacomputing/crossbow/actions/runs/6256645944/job/16987817696#step:5:569).

### DIFF
--- a/snappy.cc
+++ b/snappy.cc
@@ -74,6 +74,7 @@
 #include <cstdint>
 #include <cstdio>
 #include <cstring>
+#include <functional>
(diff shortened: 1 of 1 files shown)


danilak-G:
Fixed with https://github.com/google/snappy/commit/58d911b2841fead6ad3aedc82367ffbf3a5af8f1, will be in 1.2.2

PR #177, opened 2024-03-06
Fix the overrun issue reported by static application security testing

lian-bo:
The current issue is reported by the SAST(Static Application Security Testing) as below:

  Error: OVERRUN:
  snappy_unittest.cc:95: return_constant: Function call "sysconf(_SC_PAGESIZE)" may return -1.
  snappy_unittest.cc:95: assignment: Assigning: "page_size" = "sysconf(_SC_PAGESIZE)". The value of "page_size" is now 18446744073709551615.
  snappy_unittest.cc:97: overrun-buffer-arg: Calling "mprotect" with "this->protected_page_" and "page_size" is suspicious because of the very large index, 18446744073709551615. The index may be due to a negative parameter being interpreted as unsigned.
  #   95|       const size_t page_size = sysconf(_SC_PAGESIZE);
  #   96|       // Undo the mprotect.
  #   97|->     CHECK_EQ(0, mprotect(protected_page_, page_size, PROT_READ|PROT_WRITE));
  #   98|       CHECK_EQ(0, munmap(mem_, alloc_size_));
  #   99|     }

Let's set the page size to 4096, if the invoking sysconf(_SC_PAGESIZE) failed, otherwise still use the actual value of sysconf(_SC_PAGESIZE). In addition, also save its value in the constructor function in order to use it again in the deconstructor function, that can avoid calling the sysconf(_SC_PAGESIZE) twice.

(Did the same changes in snappy_test_tool.cc).

### DIFF
--- a/snappy_test_tool.cc
+++ b/snappy_test_tool.cc
@@ -77,11 +77,12 @@ namespace {
 class DataEndingAtUnreadablePage {
  public:
   explicit DataEndingAtUnreadablePage(const std::string& s) {
-    const size_t page_size = sysconf(_SC_PAGESIZE);
+    long page_size = sysconf(_SC_PAGESIZE);
+    page_size_ = (size_t)(page_size < 0 ? 4096 : page_size);
     const size_t size = s.size();
-    // Round up space for string to a multiple of page_size.
-    size_t space_for_string = (size + page_size - 1) & ~(page_size - 1);
-    alloc_size_ = space_for_string + page_size;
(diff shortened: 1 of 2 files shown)


lian-bo:
Thank you for applying the patch: ee6bc15edecb ("Fix the overrun issue reported by static application security testing")
So I will close this issue.

danilak-G:
Thanks for the patch. This is low priority for now as this affects only a testing tool and unittests without affecting core code

PR #187, opened 2024-06-24
Do not disable RTTI on FreeBSD since this causes other package failures

yurivict:


yurivict:
I already signed CLA.

pitrou:
Ideally RTTI should not be disabled at all: https://github.com/google/snappy/issues/189

PR #200, opened 2025-03-21
[cmake] Require CMake version 3.10 or newer

gruenich:
CMake 4.0 will be released soon and it marks
3.10 as deprecated and refuses to work with
3.5 or older versions without additional
flags for the CMake call.


gruenich:
If you consider merging this, it would be cool to create a minor release. This makes the life of packages much easier as CMake 4.0 will be released soon (don't forget to bump the library version number in the bazel file).

danilak-G:
Thank you, we will do it next week.

LGTM, I'll merge once CI is green

PR #214, opened 2025-11-10
RISC-V: gate __builtin_ctzll behind Zbb to avoid 10 % slowdown

anthony-zy:
RISC-V: use __builtin_ctzll only when Zbb is present
    
    On RV targets that implement the Zbb extension the builtin
    expands to a single `ctz` instruction and is ~10 % faster.
    Without Zbb the generic software sequence is emitted, which
    actually slows us down by the same amount.  Gate the builtin
    behind the predefined macro `__riscv_zbb` so we only use it
    when the hardware supports it.

  | 默认启用CTZ不检查zbb | 默认启用CTZ检查zbb |  
-- | -- | -- | --
编译项不带zbb | 42.5 MB/s | 47.4 MB/s | 环境不检查zbb，压缩下降10%
编译项带zbb | 50.9 MB/s | 50.9 MB/s | 环境支持zbb，压缩性能提升19.7%


danilak-G:
I think it's better to fix [CMakeLists.txt](https://github.com/google/snappy/blob/cbea40d40c61c442be7ee0c9695b45ea1b5a3c8c/CMakeLists.txt#L158) and [BUILD.bazel](https://github.com/google/snappy/blob/cbea40d40c61c442be7ee0c9695b45ea1b5a3c8c/BUILD.bazel#L146)

anthony-zy:
> I think it's better to fix [CMakeLists.txt](https://github.com/google/snappy/blob/cbea40d40c61c442be7ee0c9695b45ea1b5a3c8c/CMakeLists.txt#L158) and [BUILD.bazel](https://github.com/google/snappy/blob/cbea40d40c61c442be7ee0c9695b45ea1b5a3c8c/BUILD.bazel#L146)

Thanks for the suggestion. I've made the necessary changes to CMakeLists.txt .

PR #220, opened 2025-12-12
limit RISC-V FindMatchLength optimizations to 64-bit

yunfeizhou2025:
Add `__riscv_xlen == 64` check to ensure the optimized FindMatchLength routine is only enabled on 64-bit RISC-V platforms.


yunfeizhou2025:
Hi, @danilak-G  ,could you please review this PR when you have time? Thanks!

PR #230, opened 2026-03-11
perf: replace post-increment/decrement with pre-increment/decrement

VipulChauhan89:
### Performance Improvement: Pre-Increment/Decrement

Replaced **post-increment** (`i++`) and **post-decrement** (`i--`) operators  
with **pre-increment** (`++i`) and **pre-decrement** (`--i`) operators where applicable to improve performance.

#### Why this change matters
- Pre-increment/decrement avoids creating unnecessary temporary values.
- Helps optimize loops and frequently executed operations.
- Maintains the same logic and behavior as before.

#### Affected code
- All instances of `i++` and `i--` where pre-increment/decrement can safely be used.

### DIFF
--- a/snappy.cc
+++ b/snappy.cc
@@ -362,7 +362,7 @@ static inline bool Copy64BytesWithPatternExtension(char* dst, size_t offset) {
         // TODO: Ideally we should memset, move back once the
         // codegen issues are fixed.
         V128 pattern = V128_DupChar(dst[-1]);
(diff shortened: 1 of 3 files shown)


danilak-G:
Thanks, we are pretty sure compilers are smart enough these days to optimize it as much as they can and for integers there is no difference. If you observe performance diff, provide more details including compiler version and platform. Otherwise we will not accept this as this one is purely cosmetic

PR #235, opened 2026-04-17
RISC-V：Optimize decompression throughput by mirroring AVX fast-path for RVV short memcpy +15%

anthony-zy:
### PR Title
**Optimize RVV memcpy path to mirror AVX fast-path for short copies**

---

### Summary
Refactors the RISC-V Vector (RVV) acceleration path in the short memcpy helper to mirror the existing AVX implementation. Instead of a generic `vsetvl` loop, this version performs a fixed 32-byte vector copy (with an optional second 32-byte segment), matching the profile-driven assumption that **nearly all copies are short**.

This eliminates loop overhead in the hot path and aligns RVV behavior with the well-tuned AVX code path.

### Performance
Tested with **lzbench** on **Spacemit(R) X60** (RISC-V, RVV 1.0):

- **Before:** 269 MB/s
- **After:** 310 MB/s
- **Improvement:** **+15%**

### Implementation Details
- Uses `__riscv_vsetvl_e8m2(32)` to configure a 32-byte vector operation, matching the `kShortMemCopy` constant used in the AVX path.
- Performs an unconditional first 32-byte load/store.
- A second 32-byte segment is conditionally executed (with `SNAPPY_PREDICT_FALSE`) only when `size > kShortMemCopy`, since profiling shows long copies are rare.
- Note: with `e8m1` and `VLEN < 256` (e.g., `VLEN=128`, `vl=16`), two segments per 32 B would be required; using `e8m2` ensures 32 B per op on common configurations.

### Verification
- Environment: **Spacemit(R) X60** (RISC-V, RVV 1.0)
- Benchmark tool: **lzbench**
- Compiled with `-march=rv64gcv`
- Passed all existing Snappy unit tests.

### DIFF
--- a/snappy.cc
+++ b/snappy.cc
@@ -1244,24 +1244,22 @@ void MemCopy64(char* dst, const void* src, size_t size) {
     _mm256_storeu_si256(reinterpret_cast<__m256i *>(dst) + 1, data);
   }
   // RVV acceleration available on RISC-V when compiled with -march=rv64gcv
+  // RVV: mirror the AVX path — first 32 B, then another 32 B if size > 32.
+  // e8m1 may need two segments per 32 B when VLEN < 256 (e.g. vl=16 on VLEN=128).
 #elif defined(__riscv) && SNAPPY_HAVE_RVV
-  // Cast pointers to the type we will operate on.
-  unsigned char* dst_ptr = reinterpret_cast<unsigned char*>(dst);
-  const unsigned char* src_ptr = reinterpret_cast<const unsigned char*>(src);
-  size_t remaining_bytes = size;
-  // Loop as long as there are bytes remaining to be copied.
-  while (remaining_bytes > 0) {
-    // Set vector configuration: e8 (8-bit elements), m2 (LMUL=2).
-    // Use e8m2 configuration to maximize throughput.
-    size_t vl = VSETVL_E8M2(remaining_bytes);
-    // Load data from the current source pointer.
-    vuint8m2_t vec = VLE8_V_U8M2(src_ptr, vl);
-    // Store data to the current destination pointer.
-    VSE8_V_U8M2(dst_ptr, vec, vl);
-    // Update pointers and the remaining count.
-    src_ptr += vl;
-    dst_ptr += vl;
-    remaining_bytes -= vl;
+  assert(kShortMemCopy <= 32);
+  const size_t vl = __riscv_vsetvl_e8m2(32);
+  unsigned char* d = reinterpret_cast<unsigned char*>(dst);
+  const unsigned char* s = reinterpret_cast<const unsigned char*>(src);
+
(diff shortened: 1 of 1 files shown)


camel-cdr:
Again, how about utilizing the full vector register: https://godbolt.org/z/MovxdEzzv
Using LMUL=1 might be faster, you'll have to benchmark that.

anthony-zy:
> Again, how about utilizing the full vector register: https://godbolt.org/z/MovxdEzzv Using LMUL=1 might be faster, you'll have to benchmark that.

@camel-cdr 
"Thanks for the suggestion! I've conducted benchmarks comparing different implementations and LMUL settings. Here are the results:

Compared to the current PR, using the logic from the previous Godbolt link results in an 8% performance drop.
If I modify that Godbolt version to use LMUL=1, the performance is about 1.2% higher than the current PR.
Simply switching the current PR to LMUL=1 yields a 0.6% gain.
These results confirm that while LMUL=1 is slightly more efficient on this specific microarchitecture, the catch is that on VLEN=128bit systems, LMUL=1 only handles 16 bytes per iteration, which falls short of our 32-byte target.

To balance portability and performance, I see two options:

Option A (Current): Stick with LMUL=2. It keeps the code simple and ensures we process 32 bytes per loop on VLEN=128 systems, despite the minor 0.6% performance trade-off.
Option B (Dynamic Dispatch): Add a runtime check for vlenb. We could use LMUL=1 if vlenb >= 32 (VLEN>=256) and fallback to LMUL=2 otherwise.
I personally lean towards keeping the current LMUL=2 implementation for its robustness across different VLEN configurations, given that the 0.6% difference is marginal. What do you think?"

snappy.cc
```diff
@@ -1244,24 +1244,22 @@ void MemCopy64(char* dst, const void* src, size_t size) {
...
 #elif defined(__riscv) && SNAPPY_HAVE_RVV
-  // Cast pointers to the type we will operate on.
-  unsigned char* dst_ptr = reinterpret_cast<unsigned char*>(dst);
-  const unsigned char* src_ptr = reinterpret_cast<const unsigned char*>(src);
-  size_t remaining_bytes = size;
-  // Loop as long as there are bytes remaining to be copied.
-  while (remaining_bytes > 0) {
-    // Set vector configuration: e8 (8-bit elements), m2 (LMUL=2).
-    // Use e8m2 configuration to maximize throughput.
-    size_t vl = VSETVL_E8M2(remaining_bytes);
```
danilak-G, line 1257: I liked the macro version before, now it's not the case. If you can fix it, I'd be grateful

anthony-zy: fixed

anthony-zy: @danilak-G 
I've addressed the requested changes and updated the PR.  
If everything looks good now, could we move forward with merging?  
Thanks again for the review and support!

PR #237, opened 2026-05-01
Add NDEBUG-safe boundary checks in SnappyIOVecReader::Advance

SongT-50:
## Summary

Under `NDEBUG` (release builds) `SnappyIOVecReader::Advance()` (snappy.cc:1937) loses its
only bounds guard — a debug-only `assert(total_size_remaining_ >= curr_size_remaining_)`.
When a caller of the raw API `RawCompressFromIOVec(iov, uncompressed_length, ...)` passes an
`uncompressed_length` larger than the actual sum of `iov_len`, `Advance()` walks `curr_iov_`
past the end of the iovec array (OOB read / DoS). A run of zero-length trailing entries
triggers the same walk.

The safe public wrapper `CompressFromIOVec(iov, iov_cnt, ...)` computes the total itself and
is **unaffected** — only direct raw-API callers are exposed.

## Fix

Two NDEBUG-safe guards in `Advance()` (no `assert` reliance; API signature unchanged; normal
inputs take the same path):

1. Saturate to a no-data state and return if `total_size_remaining_ < curr_size_remaining_`
   (prevents the unsigned underflow that leads to the past-end walk).
2. Cap consecutive zero-length steps (`kMaxEmptyWalks`; `1024` is a placeholder — maintainers
   may pick a value that fits the project's convention).

## Test

Three edge cases (normal / `total_size > sum(iov_len)` / zero-length trailing) build clean under
`-fsanitize=address,undefined`; a standalone repro is available on request, and I'm happy to add
a `snappy_unittest` case.

A permanent API-level fix (add an `iov_cnt` overload and deprecate the 3-arg raw API) is the
cleaner direction — happy to follow up if you prefer that instead of / in addition to this interim guard.

### DIFF
--- a/snappy.cc
+++ b/snappy.cc
@@ -1935,13 +1935,34 @@ class SnappyIOVecReader : public Source {
  private:
   // Advances to the next nonempty `iovec` and updates related variables.
   void Advance() {
+    // Interim NDEBUG-safe guards (the assert below is stripped in release builds).
+    int empty_walks = 0;
+    constexpr int kMaxEmptyWalks = 1024;
     do {
+      // Caller-inconsistent total_size: saturate instead of unsigned-underflowing.
+      if (total_size_remaining_ < curr_size_remaining_) {
+        curr_pos_ = nullptr;
+        curr_size_remaining_ = 0;
+        total_size_remaining_ = 0;
+        return;
+      }
       assert(total_size_remaining_ >= curr_size_remaining_);
       total_size_remaining_ -= curr_size_remaining_;
       if (total_size_remaining_ == 0) {
         curr_pos_ = nullptr;
         curr_size_remaining_ = 0;
         return;
+      }
+      // Bound zero-length iovec walks (avoid spinning past the array end).
+      if (curr_size_remaining_ == 0) {
+        if (++empty_walks > kMaxEmptyWalks) {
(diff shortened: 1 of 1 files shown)


SongT-50:
Hi — a gentle ping on this PR. I know maintainers are busy, so no rush at all. If there's anything I can do to help it move forward (rebase, extra tests, or clarifying the boundary-check rationale), I'm happy to. Thanks for maintaining snappy!

SongT-50:
Hi @danilak-G — sorry to ping you directly.

I saw you reviewed and merged #242 recently, so I hope it's alright to ask here.
This PR (#237, opened May 1) touches the same file but a different function
(SnappyIOVecReader::Advance rather than MaxCompressedLength).

Status: CLA signed, check-changes passing, no conflicts.

Short version: under NDEBUG the only bounds guard in Advance() is a debug-only assert,
so a raw-API caller passing an uncompressed_length larger than the actual sum of iov_len
can walk curr_iov_ past the end of the array. CompressFromIOVec is unaffected because it
computes the total itself.

I'm not attached to the specific numbers — kMaxEmptyWalks is a placeholder and I'd happily
change the approach if a different shape fits the project better.

No rush at all, and please ignore this if it's not a priority. Thanks for maintaining snappy.

PR #242, opened 2026-06-04
Implement overflow check in MaxCompressedLength function

yusmer96-maker:
## Summary
Add integer overflow check in `MaxCompressedLength()` to prevent
heap buffer overflow vulnerabilities.

## Problem
The expression `32 + source_bytes + source_bytes / 6` can overflow
when `source_bytes` is very large, resulting in a small return value.
Callers use this value to allocate buffers, so an overflow leads to
undersized allocations and subsequent heap buffer overflows during
compression.

## Fix
Added a bounds check before the calculation. If `source_bytes` exceeds
the safe threshold `(SIZE_MAX - 32) / 7 * 6`, the function returns
`SIZE_MAX`, which forces a controlled allocation failure instead of
silent memory corruption.

### DIFF
--- a/snappy.cc
+++ b/snappy.cc
@@ -195,9 +195,9 @@ inline uint16_t* TableEntry8ByteMatch(uint16_t* table, uint64_t bytes,
 
 size_t MaxCompressedLength(size_t source_bytes) {
   // Avoid integer overflow that could cause undersized buffer allocations.
-  // Return SIZE_MAX to force a controlled allocation failure.
-  if (source_bytes > (SIZE_MAX - 32) / 7 * 6) {
-    return SIZE_MAX;
+  // Return std::numeric_limits<size_t>::max() to force a controlled allocation failure.
+  if (source_bytes > (std::numeric_limits<size_t>::max() - 32) / 7 * 6) {
+    return std::numeric_limits<size_t>::max();
   }
   // Compressed data can be defined as:
   //    compressed := item* literal*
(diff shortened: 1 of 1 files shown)


snappy.cc
```diff
@@ -194,6 +194,11 @@ inline uint16_t* TableEntry8ByteMatch(uint16_t* table, uint64_t bytes,
 }  // namespace
 
 size_t MaxCompressedLength(size_t source_bytes) {
+  // Avoid integer overflow that could cause undersized buffer allocations.
+  // Return SIZE_MAX to force a controlled allocation failure.
+  if (source_bytes > (SIZE_MAX - 32) / 7 * 6) {
```
danilak-G: LGTM. Can you use `std::numeric_limits<size_t>::max()` ?

yusmer96-maker: Sure! I initially used SIZE_MAX because I didn't notice <limits> was included transitively through snappy-stubs-internal.h. Now that I see it's available, I'll update to std::numeric_limits<size_t>::max(). Thanks for the suggestion.

snappy.cc
danilak-G, line 78: Let's have `#include <limits>` here as well

yusmer96-maker: Done, added `#include <limits>`

yusmer96-maker:
Hi @danilak-G, I've addressed all your review comments. Is there anything else needed for this to be merged? Thanks!

danilak-G:
Thanks, we are running all tests inside to verify it does not break anything. I expect this to be merged this week

PR #243, opened 2026-06-11
ci: pin action references to full commit SHAs and add read-only permissions

Alb3e3:
## Summary

- `riscv64-qemu-test.yaml` lacked a `permissions:` block entirely; added `permissions: read-all`.
- Both workflows used mutable version tags for `actions/checkout` (`@v2`, `@v4`); both are now pinned to their full commit SHAs.

Pinning prevents a compromised or rewritten upstream tag from silently injecting malicious code into CI runs.

## Verification

```
uvx zizmor --min-severity high .github/workflows/
```
Result: **no findings** after this patch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -95,7 +95,7 @@ jobs:
         matrix.optimized && 'RelWithDebInfo' || 'Debug') }}
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
       with:
         submodules: true
 
--- a/.github/workflows/riscv64-qemu-test.yaml
+++ b/.github/workflows/riscv64-qemu-test.yaml
@@ -2,6 +2,8 @@ name: riscv64-qemu-test
 
 on: [push, pull_request]
 
+permissions: read-all
+
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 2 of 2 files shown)


Alb3e3:
Two clarifications on the checks here, since this PR has been sitting since June:

**The CLA is not the blocker.** The bot comment at the top is stale — `cla/google` currently reports `✅ All contributors are covered under a CLA with Google`.

**The red check is `zizmor-output`, and none of its findings come from this PR.** All 15 are `zizmor/template-injection` at lines 104–135 of `.github/workflows/build.yml`:

```
- zizmor/template-injection at line 104 of .github/workflows/build.yml
  ... through line 135
```

Those are the pre-existing `${{ env.CMAKE_BUILD_DIR }}` / `${{ env.BINARY_PATH }}` style expansions inside `run:` blocks. This PR touches `build.yml` at one line only (line 95, an action reference) plus `riscv64-qemu-test.yaml`. zizmor scans the whole repo rather than the diff, so the check fails on any PR that touches a workflow.

I deliberately have not folded a fix for those into this PR. The remedy is to move each expansion into a step-level `env:` and reference it as a shell variable, but `build.yml` runs a Windows and Linux matrix with different shell syntax, and I cannot verify that matrix from outside the repo. Rewriting 15 expansions in a build I cannot test is exactly the kind of change that should not ride along on a small hardening PR. Happy to do it as its own PR if a maintainer wants it.

The change itself is unchanged and orthogonal: SHA-pin the action references and add read-only `permissions:` blocks.

PR #250, opened 2026-07-29
Add CompressionContext for reusing compression working memory

### LINKED ISSUE #248: Allow users to reuse working memory for `snappy` compression
Today, every call to `snappy::Compress()`/`RawCompress()` (as well as the C API `snappy_compress()`) constructs a fresh `internal::WorkingMemory` pool:

https://github.com/google/snappy/blob/1.2.1/snappy.cc#L1810

This results in a single contiguous heap allocation of around `~200 KiB` (`64KiB` hash table + `64KiB` input scratch + `MaxCompressedLength(kBlockSize)`) allocated and freed on every call. Currently, the public `snappy` API provides no way to reuse it, nor any way for users to provide their own allocator for avoiding the cost of this contiguous heap allocation.

For us at `redpanda`, this results in oversized allocation warnings, potential reactor stalls, and even potential OOMs, as we are based on the [seastar thread-per-core framework](https://github.com/scylladb/seastar) which can be quite sensitive to contiguous allocations. Indeed, our friends over at ScyllaDB (another `seastar` based project) seem to have also grappled with this:

https://github.com/scylladb/scylladb/blob/4fc12a81b3c53164d5c8290afa5a2de146f50631/transport/server.cc#L2339-L2341

Other mainstream codecs (`zstd`, `lz4`) allow for reusable contexts/working memory spaces with their public APIs, while `snappy` does not. To work around this allocation, one has to [reach into `snappy` internals and handroll their own APIs around it](https://github.com/redpanda-data/redpanda/pull/31296).

Adding some additional APIs to the public header would make this trivial, e.g.

```cpp
// snappy.h
class CompressionContext;  // reusable across calls, not thread-safe

size_t Compress(Source* reader, Sink* writer, CompressionOptions options,
                CompressionContext* ctx);
void RawCompress(const char* input, size_t input_length, char* compressed,
		 size_t* compressed_length, CompressionOptions options,
		 CompressionContext* ctx);
```

I would be happy to submit a PR upstream adding these APIs if maintainers agree this is something `snappy` should offer users!

WillemKauf:
As discussed in https://github.com/google/snappy/issues/248, `snappy` is limited today in terms of providing users a way to avoid heap allocations during compression, as well as in providing a way to reuse that scratch space across compressions.

This PR adds new APIs (while maintaining existing ABI stability) which allow users to provide their own `CompressionContext` to compression calls. This `CompressionContext` object can be used to pay the price of the heap allocation within `WorkingMemory` up-front once, and is reusable for all following compressions. A `CompressionContext` is not thread-safe, and the intended usage is one context per thread. This is largely implemented in the second commit, whereas the first commit is a simple refactoring to enable reuse of `WorkingMemory` across multiple calls to `Compress()`. In the third commit, we extend this further by allowing users to provide their own chunk of memory to `CompressionContext` using whatever allocator they see fit to avoid a raw `std::allocator<char>().allocate()` call (making use of these `snappy` APIs completely allocation free). The user is responsible for the lifetime of this memory and ensuring it is freed after the `CompressionContext` object is destroyed.

Fixes https://github.com/google/snappy/issues/248.

### DIFF
--- a/snappy-internal.h
+++ b/snappy-internal.h
@@ -137,7 +137,15 @@ inline V128 V128_DupChar(char c) { return vdupq_n_u8(c); }
 class WorkingMemory {
  public:
   explicit WorkingMemory(size_t input_size);
+
+  // Non-allocating: lays out the scratch space in the caller-provided
+  // buffer, which must be at least RequiredSize(input_size) bytes, aligned
+  // at least as strictly as uint16_t, and must outlive "*this".
+  WorkingMemory(size_t input_size, char* buffer);
   ~WorkingMemory();
+
+  // The buffer size required by the non-allocating constructor above.
+  static size_t RequiredSize(size_t input_size);
 
   // Allocates and clears a hash table using memory in "*this",
   // stores the number of buckets in "*table_size" and returns a pointer to
@@ -149,6 +157,7 @@ class WorkingMemory {
  private:
   char* mem_;        // the allocated memory, never nullptr
   size_t size_;      // the size of the allocated memory, never 0
+  bool owns_mem_;    // whether the destructor should free mem_
   uint16_t* table_;  // the pointer to the hashtable
   char* input_;      // the pointer to the input scratch buffer
   char* output_;     // the pointer to the output scratch buffer
--- a/snappy.cc
+++ b/snappy.cc
@@ -1945,6 +1945,33 @@ size_t Compress(Source* reader, Sink* writer, CompressionOptions options) {
   return InternalCompress(reader, writer, options, &wmem);
 }
 
+size_t Compress(Source* reader, Sink* writer, CompressionOptions options,
+                CompressionContext* ctx) {
+  assert(ctx != nullptr);
+  assert(ctx->working_memory_ != nullptr);
+  return InternalCompress(reader, writer, options, ctx->working_memory_);
+}
+
+CompressionContext::CompressionContext()
+    : working_memory_(new internal::WorkingMemory(kBlockSize)) {}
+
+CompressionContext::~CompressionContext() { delete working_memory_; }
+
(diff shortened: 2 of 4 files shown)


WillemKauf:
```
 /Users/runner/work/snappy/snappy/snappy-test.cc:469:52: error: variable 'dummyin' is uninitialized when passed as a const pointer argument here [-Werror,-Wuninitialized-const-pointer]
  469 |   if ( UncompressChunkOrAll(&dummyout, &dummylen, &dummyin, 0, Z_FINISH)
      |   
```

Looks like an existing snappy test is broken :slightly_smiling_face:

danilak-G:
> ```
>  /Users/runner/work/snappy/snappy/snappy-test.cc:469:52: error: variable 'dummyin' is uninitialized when passed as a const pointer argument here [-Werror,-Wuninitialized-const-pointer]
>   469 |   if ( UncompressChunkOrAll(&dummyout, &dummylen, &dummyin, 0, Z_FINISH)
>       |   
> ```
> 
> Looks like an existing snappy test is broken 🙂

Yes, new clang compiler on macos does not like something.

I'll take a look today and tomorrow and hopefully merge it without any problems. On a first pass, all looks good

WillemKauf:
> I'll take a look today and tomorrow and hopefully merge it without any problems. On a first pass, all looks good

sounds good to me, TYVM for review!

danilak-G:
LGTM, we'll run the code for a bit inside Google and make a release once we are sure everything is safe

WillemKauf:
interested to hear how the internal testing is going within Google, and when/if you're thinking of cutting another minor release soon @danilak-G! 🙂

danilak-G:
Sorry for a late response, we had pretty good internal experience and hope to cut a 1.3.0 release next week

WillemKauf:
Sweet, looking forward to it!

PR #257, opened 2026-09-06
Refuse to compress an input the stream format cannot express

lenamonj:
`InternalCompress` guarded the 2^32-byte limit with an `assert` only, so every NDEBUG build compressed a larger input into a stream whose length varint held `N mod 2^32` while the body carried all N bytes: a 4 GiB input produced a header claiming 0 bytes.

The guard is now a runtime check at the one funnel every compression entry point passes through. An input of 2^32 bytes or more writes nothing and returns zero bytes written, which no successful compression can return; `snappy_compress` returns `SNAPPY_INVALID_INPUT`. The headers document the limit. The new test drives a 4 GiB `Source` without materialising it and fails on main (201457665 bytes written).

### DIFF
--- a/snappy-c.cc
+++ b/snappy-c.cc
@@ -39,6 +39,9 @@ snappy_status snappy_compress(const char* input,
     return SNAPPY_BUFFER_TOO_SMALL;
   }
   snappy::RawCompress(input, input_length, compressed, compressed_length);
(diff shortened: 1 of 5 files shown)


lenamonj:
@googlebot I signed it!

PR #258, opened 2026-09-07
Honor CMAKE_INSTALL_INCLUDEDIR in exported header paths

### LINKED ISSUE #206: support for headers in a split prefix broken between 1.2.1 and 1.2.2
In NixPkgs/NixOS we often put headers into a different prefix than libraries. But that broke for snappy when we [attempted upgrade](https://github.com/NixOS/nixpkgs/pull/405879) from 1.2.1 to 1.2.2.

In `lib/cmake/Snappy/SnappyTargets.cmake` this update added an INTERFACE_SOURCES line that assumes everything is in a single prefix.  That's incorrect in our case and then (some) dependents would fail to find (some of) the headers.

I looked briefly into your sources and diff between the releases, but I have no idea how this difference originated.  I can (try to) provide more info in you tell me what.

Alb3e3:
Fixes #206.

The install rules honor `CMAKE_INSTALL_INCLUDEDIR`, but the exported target still
points its public header sources and include directory at `<prefix>/include`.
An external CMake consumer therefore fails to configure when headers are
installed in a separate prefix or a custom relative directory.

Use `CMAKE_INSTALL_INCLUDEDIR` for those exported paths and load GNUInstallDirs
before constructing the target. This preserves the default layout and relative
path relocation without dropping the public header sources.

Verified the original failure with an installed-package consumer. After the
change, that consumer builds and roundtrips data with default, custom relative,
absolute split-prefix, relocated default, and shared-library split-prefix
installations. All 25 unit tests and a shortened run of all benchmarks pass
with GCC 15.2 and CMake 3.31.6 on Linux.

AI assistance was used to prepare this patch and run local validation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,6 +286,9 @@ configure_file(
   "snappy-stubs-public.h.in"
   "${PROJECT_BINARY_DIR}/snappy-stubs-public.h")
 
+# Must be included before CMAKE_INSTALL_INCLUDEDIR is used.
+include(GNUInstallDirs)
+
 add_library(snappy "")
 target_sources(snappy
   PRIVATE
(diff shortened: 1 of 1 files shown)


gruenich:
Good catch!

PR #254, opened 2026-08-18
fix(cmake): resolve universal binary and multi-arch macro collision on Darwin (#223)

wittkung:
### Summary
Resolve Universal Binary (`arm64;x86_64`) SIMD feature macro collision on Apple/Darwin platforms (Fixes #223).

### Historical Context & Prior Decisions (#223)
In Issue #223 (*"Unable to create universal binaries"*), community members highlighted the difficulty of producing multi-architecture fat binaries on macOS (such as `x86_64` + `arm64` for Mac App Store distribution, universal CLI tooling, or iOS/macOS frameworks).

Historically, Snappy transitioned from Autotools to CMake, adopting `check_cxx_source_compiles()` in `CMakeLists.txt` to probe for hardware instruction extensions at configure time:
- `SNAPPY_HAVE_SSSE3` (probes `<tmmintrin.h>` for `_mm_shuffle_epi8`)
- `SNAPPY_HAVE_X86_CRC32` (probes `<immintrin.h>` for `_mm_crc32_u32`)
- `SNAPPY_HAVE_BMI2` (probes `<immintrin.h>` for `_bzhi_u32`)
- `SNAPPY_HAVE_NEON` (probes `<arm_neon.h>` for `vqtbl1q_u8`)
- `SNAPPY_HAVE_NEON_CRC32` (probes `<arm_acle.h>` for `__crc32cw`)

The configure-time results are then stamped as boolean constants into `cmake/config.h.in` via `#cmakedefine01`.

### Root Cause Analysis
When CMake is configured for Universal Binaries (e.g. `-DCMAKE_OSX_ARCHITECTURES="arm64;x86_64"`), Apple Clang is invoked with `-arch arm64 -arch x86_64` simultaneously during each `check_cxx_source_compiles()` test:

1. **The ARM64 NEON Probe**:
   - Clang attempts to compile `#include <arm_neon.h>` for both slices.
   - While valid for `arm64`, the `x86_64` slice fails compilation immediately (`fatal error: 'arm_neon.h' file not found`).
   - CMake treats the entire check as a failure -> stamps `#define SNAPPY_HAVE_NEON 0` and `#define SNAPPY_HAVE_NEON_CRC32 0`.

2. **The x86 SSSE3 / SSE4.2 / BMI2 Probes**:
   - Clang attempts to compile `#include <tmmintrin.h>` for both slices.
   - While valid for `x86_64`, the `arm64` slice fails compilation (`fatal error: 'tmmintrin.h' file not found`).
[...]

PR #261, opened 2026-09-11
CMake: Fix illegal instruction on MSVC

khuiqel:
## Problem

`CMakeLists.txt` checks `SNAPPY_HAVE_BMI2` by compiling a function with `_bzhi_u32`. On Clang and GCC, `SNAPPY_HAVE_BMI2` will be 0 if BMI2 is not enabled. On MSVC, the result will always be 1 on x86 regardless of the microarchitecture level. On CPUs without AVX2/BMI2, this will result in an illegal instruction (in `snappy::ExtractLowBytes()`). Presumably this also occurs with `SNAPPY_HAVE_SSSE3` and `SNAPPY_HAVE_X86_CRC32`, as well as `SNAPPY_HAVE_NEON_CRC32` and `SNAPPY_HAVE_NEON` on ARM, but I couldn't verify that. Can be reproduced with the `BM_UFlat` benchmarks.

The BMI2 check was added in 4f0adca, which noted the `__BMI2__` preprocessor macro doesn't exist on MSVC. Did not test MSVC+Clang.

## Solution

It's very much a band-aid, but wrap the instruction checks in `if(NOT MSVC)`. This solves the illegal instruction issue, but SSSE3/CRC32/BMI2 intrinsics will no longer generate.

```diff
+if(NOT MSVC)
+
/* ARM and other x86 intrinsics checks */

check_cxx_source_compiles("
#include <immintrin.h>
int main() {
  return _bzhi_u32(0, 1);
}" SNAPPY_HAVE_BMI2)
 
+endif(NOT MSVC)
+
```

All 25 test pass with this patch. The following tests fail without this patch (on a CPU without BMI2):

```
[  FAILED  ] 14 tests, listed below:
[  FAILED  ] CorruptedTest.VerifyCorrupted
[  FAILED  ] Snappy.SimpleTests
[  FAILED  ] Snappy.AppendSelfPatternExtensionEdgeCases
[  FAILED  ] Snappy.AppendSelfPatternExtensionEdgeCasesExhaustive
[  FAILED  ] Snappy.MaxBlowup
[  FAILED  ] Snappy.RandomData
[  FAILED  ] Snappy.CompressionContext
[  FAILED  ] Snappy.CompressionContextStaticWorkspace
[  FAILED  ] Snappy.IOVecSinkEdgeCases
[  FAILED  ] Snappy.IOVecCopyOverflow
[  FAILED  ] Snappy.ZeroOffsetCopy
[  FAILED  ] Snappy.ZeroOffsetCopyValidation
[  FAILED  ] Snappy.LiteralLengthU32Overflow
[  FAILED  ] Snappy.TestBenchmarkFiles
```

Checking for SSSE3 and SSE4.2 (x86 CRC32) support on MSVC is impossible because [MSVC only defines `__AVX__`, `__AVX2__`, and AVX-512 macros](https://learn.microsoft.com/en-us/cpp/build/reference/arch-x64?view=msvc-180) with [no plan to change](https://developercommunity.visualstudio.com/t/no-way-to-detect-ssesse2-at-compile-time-on-x64/1424304). So a proper solution for MSVC would involve adding `SNAPPY_HAVE_XXXX` options like AVX/AVX2 for SSSE3 and SSE4.2, [like Highway does](https://github.com/google/highway/commit/383e6e8931895be5d9a2db95ba6537f6d59228b7).

Somewhat related to #254


khuiqel:
Superseded by #262 as it's far less invasive

